PR #29, opened 2013-10-21
improve backend library locating/loading

dywisor:
This commit adds the `find_library` arg to the `get_backend()` functions, which allows to override the default library lookup. This is useful on systems where `ctypes.util.find_library()` cannot be used, e.g. on embedded linux where gcc and objdump are missing and the library cache is disabled (uclibc without LDSO_CACHE_SUPPORT) [0].

A (super-lazy) usage example for such systems would be

``` python
import usb.core
import usb.backend.libusb1

backend = usb.backend.libusb1.get_backend(find_library=lambda x: "/usr/lib/libusb-1.0.so")
dev     = usb.core.find(..., backend=backend)
```

Not passing `find_library` to `get_backend()` keeps the old behavior (but logging is slightly changed).

Other changes:
- libusb0, libusb1, openusb backends use the same code
  for lib loading now (usb.libloader)
- specific exceptions (derived from usb.libloader.LibraryException)
  are raised if a lib cannot be found or lib loading fails

[0] http://docs.python.org/3/library/ctypes.html#finding-shared-libraries

### DIFF
--- a/usb/__init__.py
+++ b/usb/__init__.py
@@ -48,7 +48,7 @@ version_info = (1, 0, 0, 'b1')
 __version__ = '%d.%d.%d%s' % version_info
 
 
-__all__ = ['legacy', 'core', 'backend', 'util']
+__all__ = ['legacy', 'core', 'backend', 'util', 'libloader']
 
 
 def _setup_log():
--- a/usb/backend/libusb0.py
+++ b/usb/backend/libusb0.py
@@ -27,7 +27,6 @@
 # MODIFICATIONS.
 
 from ctypes import *
-import ctypes.util
 import os
 import usb.backend
 import usb.util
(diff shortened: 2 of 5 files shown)


Boggartfly:
On Mac, which file do I load into ? The .dylib or something else?
`backend = usb.backend.libusb1.get_backend(find_library=lambda x: "/usr/share/local/lib/<insert filename with extension here>")`

walac:
I am not a Mac expert, but I think shared libraries on it have .dylib extension.

Boggartfly:
Thanks for the help, this piece of code actually saved me a lot of trouble and money! Otherwise I had to go the bootcamp route form mac.. I ended up using parallels. The native application I tried with this worked, although the windows alternative works better! Thanks a lot @walac

susamn:
@walac Works like a charm. Thanks.

balajiraju:
thanks for the post it helped me a lot

Issue #66, opened 2014-10-02
Exception when script using pyusb finishes


devbisme:
When my script executes `sys.exit()`, I get an exception:

`Exception AttributeError: "'NoneType' object has no attribute 'copy'" in <bound method Device.__del__ of <usb.core.Device object at 0x029F09D0>> ignored`

I traced this to the `release_all_interfaces()` method in core.py. It tries to copy `self._claimed_intf`, but in my case this happens to be an empty set and the exception occurs. I stopped this error by putting in a test for `len(self._claimed_intf) == 0`.

Am I making an error in how I'm using pyusb? Or is this an error in core.py?

I'm using pyusb-1.0.0b1 with python 2.7.6 and libusb-1.0 on Windows 7.

Toilal:
Thanks @walac ! Keep on the good job on this library :)

krissen:
I might misunderstand something, but the patch was merged into master, right? I've built and installed master, but on python 3.4 I'm getting the same error as reported here.

```
libusb: warning [libusb_close] internal signalling write failed, closing anyway
```

krissen:
Ping @walac (see comment above).

walac:
@krissen Sorry for the delay, can you reproduce the issue in a test script and post it here?

krissen:
No problem (re: delay)! Hmm. Unfortunately I don't write python, but I'll head over to Tigge/antfs-cli#129 and ask @Tigge if he could assist me.

PR #78, opened 2014-11-27
Finalize objects before exiting

dywisor:
This should fix issue #66:

> Starting with (C)Python 3.4, the GC features "Safe object finalization" [0],
> which doesn't work nicely with pyusb's object structure.
>
> Required globals / modules may already have been deleted (set to None)
> when an object's __del__() method gets called [1], which
> randomly breaks libusb calls etc. when calling sys.exit().
>
> Use the atexit module to solve this (indirectly via weakref.finalize()).
>
> This commit adds a class that takes care of object finalization.
> Classes/objects can adopt this functionality by inheriting
> AutoFinalizedObject and renaming their __del__() method to _finalize_object().
>
> Shouldn't break existing use cases, but adds a few layers of indirection.
> A __del__() call gets processed as follows:
> 
>   __del__() -> finalize() -> [...]
>     -> _do_finalize_object() ->  _finalize_object()
>
> All methods in this call chain are provided by AutoFinalizedObject,
> derived classes should override the no-op _finalize_object() method.
>
> The "[...]" part depends on the python version:
> * <  3.4: <nothing>
> * >= 3.4: _finalizer() -> _do_finalize_object_ref()
> 

Needs testing, esp. with Python < 2.7.

The ``FIXME DBGPRINT`` lines should be removed when merging ;)

### DIFF
--- a/usb/_objfinalizer.py
+++ b/usb/_objfinalizer.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014 André Erdmann
+#
+# The following terms apply to all files associated
+# with the software unless explicitly disclaimed in individual files.
+#
+# The authors hereby grant permission to use, copy, modify, distribute,
+# and license this software and its documentation for any purpose, provided
+# that existing copyright notices are retained in all copies and that this
+# notice is included verbatim in any distributions. No written agreement,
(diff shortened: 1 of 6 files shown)


walac:
Remove the debug print calls and merged. Thank you very much for your contribution!

PR #80, opened 2014-12-03
Add port_numbers device property

ecejeff:
Add port_numbers device property (not to be confused with existing port_number). This just wraps libusb_get_port_numbers(...) and presents the result as a tuple.

Usage: my application uses this function to find a USB device after it has re-enumerated as a bootloader with different VID and PID, but the same physical port location.

My first implementation used a string, but upon further reflection I determined that a tuple was a more natural representation, as the tuple can easily be converted to a string if necessary (and the reverse is a bit trickier).

### DIFF
--- a/tests/test_find.py
+++ b/tests/test_find.py
@@ -52,7 +52,8 @@ class _DeviceDescriptor(object):
         self.bDeviceProtocol = 0xff
         self.bus = 1
         self.address = 1
-        self.port_number= None
+        self.port_number = None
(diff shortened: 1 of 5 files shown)


walac:
Just squashed your commits and added you to the [ACKNOWLEDGEMENTS](https://github.com/walac/pyusb/blob/master/ACKNOWLEDGEMENTS#L29) file.

Thank you for your contribution.

PR #87, opened 2015-01-19
support for get usb device speed

DeliangFan:


usb/backend/libusb1.py
```diff
@@ -731,6 +731,13 @@ def get_device_descriptor(self, dev):
         except AttributeError:
             dev_desc.port_numbers = None
 
+        try:
+            dev_desc.speed = self.lib.libusb_get_device_speed(dev.devid)
+            dev_desc.iSpeed = None
```
walac: What's the role of iSpeed?

usb/backend/libusb1.py
```diff
@@ -731,6 +731,13 @@ def get_device_descriptor(self, dev):
...
+        except Exception:
```
walac: Why the exception handler here? As the call to `libusb_get_device_speed` is not wrapped by `_check`, no exception should be raised.

usb/core.py
```diff
@@ -756,6 +758,22 @@ def __init__(self, dev, backend):
...
+            if self.iSpeed == 1:
+                self.speed = 'low'
+            elif self.iSpeed == 2:
+                self.speed = 'full'
+            elif self.iSpeed == 3:
+                self.speed = 'high'
+            elif self.iSpeed == 4:
+                self.speed = 'super'
+            else:
+                self.speed = 'unknown'
```
walac: I think it is better to let only `speed` be a property and instead of hardcoded numbers, it should be use labeled constants.

walac:
The two commits can be combined in one single commit.

usb/backend/libusb1.py
```diff
@@ -731,6 +731,13 @@ def get_device_descriptor(self, dev):
         except AttributeError:
             dev_desc.port_numbers = None
 
+        try:
+            dev_desc.speed = self.lib.libusb_get_device_speed(dev.devid)
```
walac: It is a convention supply the function prototype to avoid argument type mistakes.

usb/backend/libusb1.py
DeliangFan: iSpeed represent speed type.
1: low
2: full
3: high
4: super

because libusb_get_device_speed(dev.devid) returns a int value.

usb/backend/libusb1.py
DeliangFan: Thx, you are right!

usb/core.py
DeliangFan: OK, so we only use property speed. But use what kind of labeled constants?

usb/backend/libusb1.py
walac: I think it does not have a role in the backend, does it?

usb/backend/libusb1.py
walac: Bear in mind that if `libusb_get_device_speed` returns an error, `speed` must be `None`.

usb/core.py
walac: You can define the constants:
```python
SPEED_LOW = 1
SPEED_FULL = 2
SPEED_HIGH = 3
SPEED_SUPER = 4
SPEED_UNKNOWN = -1
```

And then just assign the value:

```python
self.speed = desc.speed
```

A string representation of the speed can be achieved through the implementation of a translation function in the [`util`](https://github.com/walac/pyusb/blob/master/usb/util.py) module.

usb/backend/libusb1.py
DeliangFan: see in http://git.libusb.org/?p=libusb.git;a=blob;f=libusb/libusb.h;h=044012bcd00dbfc5f47d97745681e9e87ef775a7;hb=HEAD;js=1 
714 enum libusb_speed {
 715     /** The OS doesn't report or know the device speed. */
 716     LIBUSB_SPEED_UNKNOWN = 0,
 717 
 718     /** The device is operating at low speed (1.5MBit/s). */
 719     LIBUSB_SPEED_LOW = 1,
 720 
 721     /** The device is operating at full speed (12MBit/s). */
 722     LIBUSB_SPEED_FULL = 2,
 723 
 724     /** The device is operating at high speed (480MBit/s). */
 725     LIBUSB_SPEED_HIGH = 3,
 726 
 727     /** The device is operating at super speed (5000MBit/s). */
 728     LIBUSB_SPEED_SUPER = 4,
 729 };

usb/core.py
DeliangFan: let  SPEED_UNKNOWN = 0

usb/backend/libusb1.py
walac: hrm I didn't get it yet, `iSpeed` still looks useless in the backend for me.

usb/core.py
walac: Sure :)

PR #88, opened 2015-01-27
Support for getting usb device speed

DeliangFan:
In some cases, we need to get the usb device speed type to attach
it to specific usb controller. Fortunately, it's easy to get usb
device speed information with libusb1.0.


walac:
lgtm, merged. Thanks.

PR #91, opened 2015-02-10
util: change tostring() to tobytes() when running with Python 3.2+

tuzzer:
This removes the DeprecationWarning that warns that tostring() is obsolete when running with Python 3.2+


walac:
Merged, thanks.

PR #93, opened 2015-02-24
Claim interface only for standard control transfer requests.

damib:
wIndex is the interface number to be claimed only for
standard request types. For example Silicon Labs cp2103 defines
a vendor control transfer to set some GPIO pins though wIndex
parameter (se AN571 page 17).


walac:
Merged, thanks.

PR #97, opened 2015-04-26
Support the libusb0 that ships in Solaris

jcflack:
Solaris ships with a libusb0 implementation (that works both with usb devices attached to the host, and those attached to SunRay thin clients served by the host!). But the libusb0 backend wasn't working right because the wrong value was being obtained for PC_PATH_MAX. Fixed here.

One problem remains: the library on Solaris is found at /usr/sfw/lib/libusb.so.1 which is not found automatically by the backend. I am not sure what would be your preferred way to fix that. Just add to the list of names in _load_library? Define a different _load_library if sys.platform is sunos5? Something else?

For right now, the client code in my application is calling get_backend directly to handle this problem, but it will be nice to take that out once the backend can find the library without help.

### DIFF
--- a/usb/backend/libusb0.py
+++ b/usb/backend/libusb0.py
@@ -46,7 +46,7 @@ _logger = logging.getLogger('usb.backend.libusb0')
 # usb.h
 
 if sys.platform.find('bsd') != -1 or sys.platform.find('mac') != -1 or \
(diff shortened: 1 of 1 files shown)


walac:
Merged, thanks.

PR #98, opened 2015-05-03
Conform _interop to the README goal to support 2.4+.

jcflack:
README.rst was saying 2.4+ should be supported but some vestiges from 2.3
were in _interop.py.

http://sourceforge.net/p/pyusb/mailman/message/34078264/

### DIFF
--- a/usb/_interop.py
+++ b/usb/_interop.py
@@ -34,10 +34,10 @@
 import sys
 import array
 
-__all__ = ['_reduce', '_set', '_next', '_groupby', '_sorted', '_update_wrapper']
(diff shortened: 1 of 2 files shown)


walac:
Since we are refactoring things, we should remove `_sorted` and `_groupby` and replace them by their builtin counterparts.

jcflack:
I'm not sure if you get notified when I add a commit, or only when I add a comment.

So, here's a comment saying I added a commit. :)

walac:
lgtm. Just squash both commits (`git rebase -i` and forced push do the job).

jcflack:
Squashed.

walac:
Merged, thanks.

PR #99, opened 2015-05-03
Handling device strings; making device's LANGID list accessible.

jcflack:
This started off just as an exercise in not having the serial number come back as Љ if the device doesn't have one, and grew into a little more work to expose the device's available language IDs as a property on Device.

### DIFF
--- a/usb/core.py
+++ b/usb/core.py
@@ -720,6 +720,7 @@ class Device(_objfinalizer.AutoFinalizedObject):
         self._ctx = _ResourceManager(dev, backend)
         self.__default_timeout = _DEFAULT_TIMEOUT
         self._serial_number, self._product, self._manufacturer = None, None, None
+        self._langids = None
 
         desc = backend.get_device_descriptor(dev)
 
@@ -768,6 +769,25 @@ class Device(_objfinalizer.AutoFinalizedObject):
             self.speed = int(desc.speed)
         else:
             self.speed = None
+
+    @property
+    def langids(self):
+        """ Return the USB device's supported language ID codes.
+
+        These are 16-bit codes familiar to Windows developers, where for
+        example instead of en-US you say 0x0409. USB_LANGIDS.pdf on the usb.org
+        developer site for more info. String requests using a LANGID not
+        in this array should not be sent to the device.
+
+        This property will cause some USB traffic the first time it is accessed
+        and cache the resulting value for future use.
+        """
+        if self._langids is None:
+            try:
+                self._langids = util.get_langids(self)
+            except USBError:
+                self._langids = ()
+        return self._langids
 
     @property
     def serial_number(self):
--- a/usb/util.py
+++ b/usb/util.py
@@ -39,6 +39,7 @@ find_descriptor - find an inner descriptor.
 claim_interface - explicitly claim an interface.
 release_interface - explicitly release an interface.
 dispose_resources - release internal resources allocated by the object.
+get_langids - retrieve the list of supported string languages from the device.
 get_string - retrieve a string descriptor from the device.
 """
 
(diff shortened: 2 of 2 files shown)


usb/core.py
```diff
@@ -770,13 +771,32 @@ def __init__(self, dev, backend):
...
+        in this array should not be sent to the device.
+
+        This property will cause some USB traffic the first time it is accessed
+        and cache the resulting value for future use.
+        """
+        if self._langids is None:
+            try:
+                self._langids = util.get_langids(self)
+            except USBError:
+                self._langids = array.array('H')
```
walac: Why this?

usb/core.py
```diff
@@ -770,13 +771,32 @@ def __init__(self, dev, backend):
...
+
+    @property
     def serial_number(self):
         """ Return the USB device's serial number string descriptor.
 
         This property will cause some USB traffic the first time it is accessed
         and cache the resulting value for future use.
         """
-        if self._serial_number is None:
+        if self._serial_number is None and self.iSerialNumber > 0:
```
walac: Is this to prevent an exception in case there is no serial number string?

usb/core.py
```diff
@@ -787,7 +807,7 @@ def product(self):
         This property will cause some USB traffic the first time it is accessed
         and cache the resulting value for future use.
         """
-        if self._product is None:
+        if self._product is None and self.iProduct > 0:
```
walac: Same as `iSerialNumber`?

usb/core.py
```diff
@@ -798,7 +818,7 @@ def manufacturer(self):
         This property will cause some USB traffic the first time it is accessed
         and cache the resulting value for future use.
         """
-        if self._manufacturer is None:
+        if self._manufacturer is None and self.iManufacturer > 0:
```
walac: Same as `iSerialNumber`?

usb/util.py
```diff
@@ -241,6 +242,53 @@ def dispose_resources(device):
...
+    if len(buf) < 4 or buf[0] < 4:
+        return langids
+
+    buf = buf[2:buf[0]]
+    buf.extend((0, 1)) # in proper endianness this should represent 256
+
+    if hexversion >= 0x03020000:
+        langids.frombytes(buf.tobytes())
+    else:
+        langids.fromstring(buf.tostring())
```
walac: `get_descriptor` returns an `array.array` object, why all these conversions are necessary?

usb/util.py
```diff
@@ -254,22 +302,15 @@ def get_string(dev, index, langid = None):
     The return value is the unicode string present in the descriptor.
     """
     from usb.control import get_descriptor
+    langids = dev.langids
+
+    if 0 == len(langids):
+        raise ValueError("get_string from a device that offers none")
```
walac: All think this can be removed. If an app wants to make sure the langid exists, if can do it by itself.

usb/util.py
```diff
@@ -254,22 +302,15 @@ def get_string(dev, index, langid = None):
...
-                    254,
-                    DESC_TYPE_STRING,
-                    0
-                )
-        assert len(buf) >= 4
-        langid = buf[2] | (buf[3] << 8)
+        langid = langids[0]
+    else:
+        if langid not in langids:
+            raise ValueError("get_string in a language device does not support")
```
walac: This change can also be reverted.

usb/core.py
jcflack: Sorry ... I didn't squash these 4 commits because I thought the commit messages in order would help make the intention clear. This is from the fourth commit, handling the fact that some devices with no LANGIDs don't even return an empty LANGIDs array, they return _nothing_ and a backend pipe error results. Client code should treat the case the same as no LANGIDs supported, best achieved by returning the same thing (empty array) to client code in either case.

usb/core.py
jcflack: Not an exception but a bogus string like Љ (resulting from misinterpretation of the LANGIDs array at index zero as if it were a utf-16-le string). Clear in the USB standard that using index zero for a string indicates there is no string.

usb/core.py
jcflack: Yes.

usb/core.py
jcflack: Yes.

usb/util.py
jcflack: Here was my reasoning: USB spec explicitly requires that no get string descriptor request ever be sent to a device for a langid the device has not offered. If that is left for every app that "wants" to make sure, then every app had better "want" to make sure, or noncompliant USB traffic will be sent. If checked here, then pyusb avoids generating noncompliant traffic, and if any client code is making a buggy assumption or failing to check, it will be caught here.

usb/util.py
jcflack: I assume you just mean the test and exception for unsupported langid. The rest of the change goes hand in hand with factoring out the langid retrieval so it is accessible to client code, so it will be possible to write correct clients.

usb/util.py
jcflack: get_descriptor returns an array.array('B') object, but LANGIDs are short ints and should be presented that way to the client. Original array.array seems not to have an easy cast to a new shape method as Python 3 memoryviews do. However, as I think more, there is no need for this to be fancy at all. LANGIDs could be returned as a plain Python tuple of ints. Simpler all around. Absent any objection, I'll change it to that.

jcflack:
Changed the langids property from array.array('H') to simple tuple of Python ints.
Code ends up simpler.

usb/util.py
walac: Sounds reasonable. Please, just replace the error string by "The device has no langid".

usb/util.py
walac: This was based on my earlier comment above, but now it is irrelevant. Just replace the error string by "The device does not support the specified langid".

usb/util.py
```diff
@@ -241,6 +242,40 @@ def dispose_resources(device):
...
+                dev,
+                254,
+                DESC_TYPE_STRING,
+                0
+            )
+
+    if len(buf) < 4 or buf[0] < 4 or buf[0]&1 != 0:
+        return ()
+
+    return tuple(map(lambda x,y: x+(y<<8), buf[2:buf[0]:2], buf[3:buf[0]:2]))
```
walac: Would be nice if you put a small comment of what is going on here :)

walac:
There is just one small nit, then just squash all but the first commit and it is ready to merge.

jcflack:
Didn't squash yet ... this commit improves some comments, also moves the test for index == 0 into get_string itself, as zero for a string index always means the device isn't supplying a string. That way, I can revert the changes from the first commit, ironically enough; the call sites don't all have to check for index zero if get_string is doing it.

walac:
Nice catch on moving the zero check to the call site :) I didn't review docs yet, once you squash the commits, I make the review, and if it is ok, I merge the PR. Thanks for working on this.

walac:
Merged, thanks.

PR #103, opened 2015-05-31
fix 0.x compatibility issue in functions setConfiguration() and setAltInterface()

AndreySV:
Please look at the patches to be sure, that this is a proper solution for the problem, described in the commit messages. These changes helped to run working script for 0.x API with the latest pyusb without any changes.
These patches were tested on Windows and on Linux.

### DIFF
--- a/usb/legacy.py
+++ b/usb/legacy.py
@@ -256,7 +256,7 @@ class DeviceHandle(_objfinalizer.AutoFinalizedObject):
         Arguments:
             configuration: a configuration value or a Configuration object.
         """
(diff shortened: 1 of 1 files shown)


walac:
Indeed it is bug, but the methods also should accept if the proper values are passed, i.e, `configuration.value` and `alternate.alternateSetting`.

AndreySV:
Is it ok, if I fix this like it's done in legacy claimInterface function?

        if isinstance(interface, Interface):
            if_num = interface.interfaceNumber
        else:
            if_num = interface

PR #104, opened 2015-05-31
fix 0.x compatibility issue in functions setConfiguration() and setAltInterface()

AndreySV:
Updated patches, that we have discussed here: https://github.com/walac/pyusb/pull/103

### DIFF
--- a/usb/legacy.py
+++ b/usb/legacy.py
@@ -303,6 +303,9 @@ class DeviceHandle(_objfinalizer.AutoFinalizedObject):
         Arguments:
             interface: interface number or an Interface object.
         """
+        if isinstance(interface, Interface):
+            interface = interface.interfaceNumber
+
         self.dev.detach_kernel_driver(interface)
 
 class Device(object):
--- a/usb/legacy.py
+++ b/usb/legacy.py
@@ -267,6 +267,9 @@ class DeviceHandle(_objfinalizer.AutoFinalizedObject):
         Arguments:
             alternate: an alternate setting number or an Interface object.
         """
+        if isinstance(alternate, Interface):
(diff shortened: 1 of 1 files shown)


usb/legacy.py
```diff
@@ -256,7 +256,13 @@ def setConfiguration(self, configuration):
         Arguments:
             configuration: a configuration value or a Configuration object.
         """
-        self.dev.set_configuration(configuration)
+        if isinstance(configuration, Configuration):
+           conf_num = configuration.value
+        else:
+           conf_num = configuration
+
+        self.dev.set_configuration(conf_num)
```
walac: hrm I think this is better:

```python
if isinstance(configuration, Configuration):
    configuration = configuration.value

self.dev.set_configuration(configuration)
```

usb/legacy.py
```diff
@@ -264,7 +270,12 @@ def setAltInterface(self, alternate):
         Arguments:
             alternate: an alternate setting number or an Interface object.
         """
-        self.dev.set_interface_altsetting(self.__claimed_interface, alternate)
+        if isinstance(alternate, Interface):
+           alt_num = alternate.alternateSetting
+        else:
+           alt_num = alternate
+
+        self.dev.set_interface_altsetting(self.__claimed_interface, alt_num)
```
walac: The same as for configuration applies here.

walac:
Tip: you don't need to open a num PR, just make the fixes, squash de commits and do a force push that this PR will be updated.

AndreySV:
I fixed patches according to your comments. Additionaly I fixed similar problem in detachKernelDriver and reworked a little bit claimInterface.

walac:
Merged, thank you.

PR #105, opened 2015-05-31
For discussion anyway, one attempt at some relevant trove classifiers.

jcflack:
These are the ones I came up with by reading down the PyPi trove classifier list while thinking "who would be using a usb library for what?". For the operating systems, I looked around for what was said to support any of the backend libraries, and for Python implementations I relied on a paper naming several that either did or intended to (in 2006) support ctypes.

I thought about a bunch of the multimedia classifiers (sound and video capture, MIDI, etc.) as a lot of that stuff is usb, but decided to leave them out because those classifier might be more appropriate for specific  code working on top of pyusb.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,42 @@ PyUSB offers easy USB devices communication in Python.
 It should work without additional code in any environment with
 Python >= 2.4, ctypes and an pre-built usb backend library
 (currently, libusb 0.1.x, libusb 1.x, and OpenUSB).
-"""
+""",
+    classifiers=[
+        'Development Status :: 4 - Beta',
+        'Intended Audience :: Developers',
+        'Intended Audience :: Information Technology',
+        'Intended Audience :: Manufacturing', # USB automation, or mfg USB devs
+        'Intended Audience :: Science/Research', # interface with instruments
+        'Intended Audience :: System Administrators', # integrate strange devs
+        'Intended Audience :: Telecommunications Industry', # telecomm devs
+        'License :: OSI Approved :: BSD License',
+        'Natural Language :: English',
+        # try to union the OSes that can build any of the backend libraries...
+        'Operating System :: MacOS :: MacOS X',
+        'Operating System :: Microsoft :: Windows :: Windows Vista',
+        'Operating System :: Microsoft :: Windows :: Windows 7',
+        'Operating System :: POSIX :: BSD :: FreeBSD',
+        'Operating System :: POSIX :: BSD :: NetBSD',
+        'Operating System :: POSIX :: BSD :: OpenBSD',
+        'Operating System :: POSIX :: Linux',
+        'Operating System :: POSIX :: SunOS/Solaris',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -48,6 +48,42 @@
 It should work without additional code in any environment with
 Python >= 2.4, ctypes and an pre-built usb backend library
 (currently, libusb 0.1.x, libusb 1.x, and OpenUSB).
-"""
+""",
+    classifiers=[
+        'Development Status :: 4 - Beta',
```
walac: Next version is supposed to be [release candidate](https://github.com/walac/pyusb/blob/master/usb/__init__.py#L49). Moreover, is there a way to avoid duplicate info with `__init__.py`?

setup.py
```diff
@@ -48,6 +48,42 @@
...
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
+        # source(CPython,Jython,IronPython,PyPy): "The Long Term" section of
+        # http://ojs.pythonpapers.org/index.php/tpp/article/viewFile/23/23
+        'Programming Language :: Python :: Implementation :: CPython',
+        'Programming Language :: Python :: Implementation :: IronPython',
+        'Programming Language :: Python :: Implementation :: Jython',
+        'Programming Language :: Python :: Implementation :: PyPy',
+        'Topic :: Scientific/Engineering :' \
+            ': Interface Engine/Protocol Translator',
```
walac: I think this can go on one line.

setup.py
jcflack: In the PyPi trove classifier scheme, the next status up (5) is "Production/Stable". I can see that the mapping between the semver.org scheme and the trove scheme might not be exactly 1-to-1, but there's probably a mechanical way to do it. I would think that anything _up to_ a 1.x release in the semver scheme, including a candidate, would still be fit into 'beta' in the trove scheme (unless the semver tag is -alpha* or something).  I'm not sure how that would apply to, say, 1.3.0-rc1, but that's a ways off....

I've looked around at techniques for reducing the redundancy - many seem to recommend that setup.py should avoid actually importing the module, but should simply parse either `__init.py__` or a dedicated version file for the version information only. I wasn't sure what you'd want to do there.

setup.py
jcflack: I can see this'll come down to whether you hate having it line-broken more than I hate making >80-column text files. :) Maybe we can sell tickets as a fundraiser....

setup.py
walac: Seems that beta is the best fit here, as there is no frozen/rc status in trove.

Regarding the duplicate information, I was just predicting me updating one file and forgetting to update the other.

setup.py
walac: lol In general I try to keep my lines < 80-column, but I am not all crazy about that. This is just a nit, however, if you feel better is should stay this way, no problem.

walac:
Merged, thanks.

PR #109, opened 2015-08-16
fix get_device_speed for old libusb versions

AlanJAS:
This works for me. Tested on Ubuntu 10.04 (Lucid) with libusb-1.0-0 package version 2:1.0.6-1


walac:
Didn't you get an exception [here](https://github.com/walac/pyusb/blob/master/usb/backend/libusb1.py#L557)?

AlanJAS:
No. Only when try call the function lib.libusb_get_device_speed.

walac:
Merged, thanks.

PR #113, opened 2015-08-26
prevent get_device_speed error on libusb1

AlanJAS:
Fix possible problem with previous request.


walac:
I suspected this you happen, thanks.

AlanJAS:
Thanks! I'm not sure about use try/except in this cases. The correct way must be check with hasattr. In libusb0.py appears similar checks.

PR #119, opened 2015-11-23
Fix OpenUSB link

M3gaFr3ak:
The link for OpenUSB doesn't work. It says "Page not found" on sourceforge. I replaced it with the link to the OpenUSB sourceforge wiki


walac:
Thanks.

PR #124, opened 2016-01-08
ctrl_transfer: claim receiving interface for non-VENDOR transfers

roques:
This matches what the Linux kernel does in `check_intf()` when called from `check_ctrlrecip()`
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/drivers/usb/core/devio.c
and thus helps to prevent annoying warnings from the kernel.

The previous code did this for `STANDARD` type transfers already, but not for `CLASS`.


walac:
Thank you.

PR #125, opened 2016-01-14
make find and find_descriptor easier to understand

swinman:
all( iter_ ) is more explicit than reduce( lambda x, y:x and y, iter_, True ),
especially since reduce is no longer a builtin.  Use all instead of
complicated multi-line reduce statement for easier code comprehension.
Also, introduce _all interop since all was introduced in 2.5


walac:
Very good. Thanks :)

PR #137, opened 2016-03-17
Typo fix: replace LibaryException with LibraryException.

ixjlyons:
I actually realize you might want to keep this typo to prevent a breaking change, but it's an easy fix I thought I'd offer.


walac:
Thanks, I'll merge it after stable release.

PR #141, opened 2016-05-01
Fix setup.py for v1.0.0

eblot:
I think the setup.py file needs to be updated for the final release of v1.0.0.
Cheers.


walac:
Thanks :)

PR #146, opened 2016-05-19
Quick Debian packaging

ehem:
Installing yet another tool usually isn't too difficult a task, but
tracking them and very importantly removing them later becomes a PITA
unless the package manager knows about them.  This is a very quick
packaging job for Debian, which I hope works for others, but YMMV.
Simply run `dpkg-buildpackage -b` and install the result.


walac:
Thanks.

tormodvolden:
This is not a good idea. pyusb is already packaged by Debian. Having debian folders in upstream tarballs is frowned upon by Debian maintainers. It means they will have to repack the release tarball to remove this folder (and Debian cares a lot about the integrity of the upstream tarball). So if you want to include custom Debian packaging, "hide" it in a subfolder or call it something else, with instructions to how it can be used. 

Furthermore, the resulting version in this "fake" package doesn't reflect that this is a DIY package, and the user might get confused later, or miss official Debian updates. It should have a version "number" (could include letters and ~) to make it stand out when listed, and also have a low  enough version so that official updates still are offered by the package managers.

ehem:
Gah!  I **deliberately looked** for such an existing package and didn't see it.  They did it using the names "python-usb" and "python-usb3" and I was looking for something closer to "pyusb".  Alas that is likely a better packaging job (mine was a quick job that worked and readily allowed removal afterwords).  Yuck.

tormodvolden:
apt-cache search pyusb :)

walac:
@tormodvolden hrm, I was not aware Debian team used a subfolder with the same name. I am in a trip and will revert the commit when I am back. Thanks for pointing this out.

PR #152, opened 2016-08-12
legacy: set Bus.location and Device.devnum

BinaryMan32:
Properly implement legacy compatibility for the following fields which
were previously set to default values:

 Bus.location = 0
 Device.devnum = None


walac:
Tks.

PR #167, opened 2017-01-04
enable building as a wheel by using setuptools

James-Dengel:
This enables building the package as a wheel.


udifuchs:
Could you upload the wheel to pypi?
I am using pynsist that only works with wheel files.

Issue #180, opened 2017-05-04
Building with recent buildout always breaks the first time because of capitals in project name


StefanRijnhart:
Building PyUSB from buildout 2.9.0 or higher breaks the first time. After that, buildout finishes completely. The problem seems to be a confusion between __pyusb__ and __PyUSB__ as project names.

```
Getting distribution for 'pyusb>=1.0.0b1'.
zip_safe flag not set; analyzing archive contents...
While:
  Installing odoo.
  Getting distribution for 'pyusb>=1.0.0b1'.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/buildout/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/buildout.py", line 1982, in main
    getattr(buildout, command)(args)
  File "/home/buildout/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/buildout.py", line 668, in install
    installed_files = self[part]._call(recipe.install)
  File "/home/buildout/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/buildout.py", line 1419, in _call
    return f()
  File "/home/buildout/anybox.recipe.odoo/anybox/recipe/odoo/base.py", line 1114, in install
    self.install_requirements()
  File "/home/buildout/anybox.recipe.odoo/anybox/recipe/odoo/base.py", line 557, in install_requirements
    eggs_recipe.install()
  File "/home/buildout/eggs/zc.recipe.egg-2.0.3-py2.7.egg/zc/recipe/egg/egg.py", line 126, in install
    reqs, ws = self.working_set()
  File "/home/buildout/eggs/zc.recipe.egg-2.0.3-py2.7.egg/zc/recipe/egg/egg.py", line 84, in working_set
    allow_hosts=self.allow_hosts)
  File "/home/buildout/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 914, in install
    return installer.install(specs, working_set)
  File "/home/buildout/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 666, in install
    for dist in self._get_dist(requirement, ws):
  File "/home/buildout/eggs/zc.buildout-2.9.0-py2.7.egg/zc/buildout/easy_install.py", line 564, in _get_dist
    dists = [_move_to_eggs_dir_and_compile(dist, self._dest)]
[...]

PR #189, opened 2017-08-18
[ADD] pyusb module to satisfy setuptools

hbrunn:


rnixx:
I don't understand what this change solves. @hbrunn what's the issue here?

hbrunn:
this solves #180 because now, we have a module with the name expected by setuptools

rnixx:
@hbrunn sorry, that's not true. the problem in #180 was caused by #192. See https://github.com/walac/pyusb/issues/180#issuecomment-335115008

rnixx:
@walac imho this change can be reverted, no need to import from "pyusb", which is what this change enables.

hbrunn:
well, I don't care one way or another. This commit enabled me to just dump the url to this branch into a buildout and this works, if whatever you did has the same effect, I'm just as happy

walac:
I merged this PR by mistake, I already reverted it but just didn't push it.

Issue #154, opened 2016-08-17
UnicodeDecodeError


RWAP:
I have connected a USB receipt printer to a BananaPi and then use Python to read the descriptors for the printer.

The code contains the following lines:

```
callString = callString + 'idVendor=' + padded_hex(int(cfg.idVendor),4) + ' ' + \
    'idProduct=' + padded_hex(int(cfg.idProduct),4) + ' ' + \
    'bcdDevice=' + str(cfg.bcdDevice) + ' ' + \
    'iManufacturer="' + str(usb.util.get_string(cfg,cfg.iManufacturer) ) + '" ' + \
    'iProduct="' + str(usb.util.get_string(cfg,cfg.iProduct) ) + '" ' + \
    'iSerialNum="' + str(usb.util.get_string(cfg,cfg.iSerialNumber) ) + '" ' +\
    'IPNPstring="'. + IPNPstring + '"'
```

Whilst my code works fine with a HP printer, with the USB receipt printer I get the error:

File "/g_printer_define.py", line 90, in <module>
 'IPNPstring="'. + IPNPstring + '"'
File "/usr/local/lib/python2.7/dist-packages/usb/util.py", line 330, in get_string
  return buf[2:buf[0]].tostring().decode('utf-16-le')
File "/usr/lib/python2.7/encodings/utf_16_le.py", line 16, in decode
  return codecs.utf+16+le+decode(input, errors, True)
UnicodeDecodeError: 'utf16' codec can't decode byte 0x31 in position 26: truncated data

I have asked the printer manufacturer for the details which their device passes for those strings, but should have thought that pyusb should be able to handle any characters permitted by the USB protocol?

RWAP:
The lsusb -v output is:

```
Bus 004 Device 002: ID 154f:0517 SNBC CO., Ltd 
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 (Defined at Interface level)
  bDeviceSubClass         0 
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x154f SNBC CO., Ltd
  idProduct          0x0517 
  bcdDevice            1.f0
  iManufacturer           1 BEIYANG
  iProduct                2 BTP-R580(U)
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength           39
    bNumInterfaces          1
[...]

PR #204, opened 2018-07-13
Handle missing null terminator situation for device descriptors

bgruberfossil:
I have found that certain device vendors provide invalid device descriptors that are missing a null terminator. In that case, this library crashes during the call to decode.

This provides a quick fix for that situation and allows our software, which uses pyusb, to function properly.

### DIFF
--- a/usb/util.py
+++ b/usb/util.py
@@ -311,7 +311,10 @@ def get_string(dev, index, langid = None):
                 index,
                 langid
             )
+    part = buf[2:buf[0]]
+        if part[-1] != 0:
+            part.append(0)
     if hexversion >= 0x03020000:
-        return buf[2:buf[0]].tobytes().decode('utf-16-le')
+        return part.tobytes().decode('utf-16-le')
     else:
-        return buf[2:buf[0]].tostring().decode('utf-16-le')
+        return part.tostring().decode('utf-16-le')


jonasmalacofilho:
The USB spec explicitly states that.

> The UNICODE string descriptor (shown in Table 9-16) is not NULL-terminated. The string length is computed by subtracting two from the value of the first byte of the descriptor.
—Universal Serial Bus Specification Revision 2.0, (9.6.7) String

However, while many valid UTF-16-LE strings will end with a `0x00` byte, that's just actually part of the last code point

```
>>> bytes('D', 'utf-16-le').hex()
'4400'
```

and the following is also perfectly valid in Python 3:

```
>>> bytes('Δ', 'utf-16-le').hex()
'9403'
>>> bytes.fromhex('9403').decode('utf-16-le')
Δ
```

So by introducing an extra `0x00` byte these changes can _actually break_ the decoding of valid strings that end in a code point > 255:

```
>>> bytes.fromhex('940300').decode('utf-16-le')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.7/encodings/utf_16_le.py", line 16, in decode
    return codecs.utf_16_le_decode(input, errors, True)
UnicodeDecodeError: 'utf-16-le' codec can't decode byte 0x00 in position 2: truncated data
```

mcuee:
So the pull request is in reality not valid.

jonasmalacofilho:
@bgruberfossil your issue is probably related to #154 ("UnicodeDecodeError").

PR #205, opened 2018-07-19
miss me with that weeb shit

jaredwelsh:


ruffm:
I agree. This would be very a useful change.

adanchak36:
I feel like the program is more fluid with this change.

Issue #199, opened 2018-05-04
Feature: usb.control.get_status() does not have any option to get non-standard descriptors


Jude188:
I am working with USB Hubs and am wanting to get usb hub's 'hub descriptor' as specified in the USB spec. I cannot just use:
```python
usb.control.get_status(hub, 1024, usb.DT_HUB, 0)
```
because the bmRequestType is hard coded to be generated with:
```python
bmRequestType = usb.util.build_request_type(
                    util.CTRL_IN,
                    util.CTRL_TYPE_STANDARD,
                    util.CTRL_RECIPIENT_DEVICE)
```
The util.CTRL_TYPE_STANDARD is the problem here. I need to use util.CTRL_TYPE_CLASS. This seems like a feature which could be fairly easily added and I am happy to look at this myself if you think that it would be a good addition. However, I thought it would be good to start a discussion about how to deal with the issue before starting. Should there be an additional argument to determine which CTRL_TYPE to use?   

For the moment my solution is to generate the bmRequestType myself and use
```python
bmRequestType = usb.util.build_request_type(
                    usb.util.CTRL_IN,
                    usb.util.CTRL_TYPE_CLASS,
                    usb.util.CTRL_RECIPIENT_DEVICE)

wValue = usb.DT_HUB << 8

dev.ctrl_transfer(bmRequestType, usb.REQ_GET_DESCRIPTOR, wValue=wValue, wIndex=0, data_or_wLength=1024)
```

I'm sure that this issue probably spans across at least the rest of the descriptor functions in usb.control and possibly others too.

Any feedback or comments would be welcomed! I've really only just started using pyusb!

mcuee:
Seems to be a valid request and we can add a additional entry in the pyusb/control.py file.

Jude188:
When you say an additional entry do you mean an additional argument? Because looking at this (https://www.beyondlogic.org/usbnutshell/usb6.shtml) it seems like really CTRL_TYPE_STANDARD should not be hardcoded anywhere. If this added as a keyword argument with a default value of CTRL_TYPE_STANDARD it should't result in a breaking change. 

Though that is a little ugly. Doesn't seem worth introducing a breaking change though :/

Anyway, this is just my thoughts having not looked or thought about this for about a year and a half!

Jude188:
Please do let me know if you agree that this issue is more widespread than just descriptors in the control functions.

jonasmalacofilho:
@jude188

[...]

PR #282, opened 2019-12-07
Implements #199: Remove hardcoded request_type for getting and setting descriptors

Jude188:
Issue #199

### DIFF
--- a/usb/control.py
+++ b/usb/control.py
@@ -132,7 +132,7 @@ def set_feature(dev, feature, recipient = None):
                       wIndex = wIndex,
                       wValue = feature)
 
-def get_descriptor(dev, desc_size, desc_type, desc_index, wIndex = 0):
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
By the way, thank you @Jude188!

And I'm sorry it took the original issue didn't get a response sooner.  But we now have new maintainers (I'm the newest guy), and this should improve.

Also, you raised an important point when you suggested that we should think about  other class-specific requests (IMO: as long as they are part of the spec), which is why my previous comment focused on a more general solution.

jonasmalacofilho:
Overloading the methods for fetching _standard device requests_ is, IMO, not the best solution.

> I think these and the other hub class-specific requests outlined in #199 should be implemented as new methods in the control module.
>
> This would allow properly validating that dev is indeed of class hub, as well as providing sensible documentation.

No counter-arguments have been made, and I don't plan to merge the PR as is without being convinced that it's indeed the most sensible solution.  So I'm closing this (for now, at least).

Regardless, thanks for proposing this approach and bringing the issue to my attention.

Issue #285, opened 2019-12-28
Termux (Android) support


Querela:
Hi,

In Termux you only receive a USB file descriptor (number bound to single process).
Any operations like `usb.core.find(..)` won't work.
I wrote a workaround/fix that extends the libusb1 interface and provides the `Device` object given the file descriptor.
See here: https://github.com/Querela/termux-usb-python/blob/e6166ded4555d988365e6609b91dec4f5ebb8fcd/usblib.py#L94

```python
def device_from_fd(fd):
    # setup library
    backend = libusb1.get_backend()
    lib = backend.lib
    ctx = backend.ctx

    # extend c wrapper with android functionality
    lib.libusb_wrap_sys_device.argtypes = [
        libusb1.c_void_p,
        libusb1.c_int,
        libusb1.POINTER(libusb1._libusb_device_handle),
    ]

    lib.libusb_get_device.argtypes = [libusb1.c_void_p]
    lib.libusb_get_device.restype = libusb1._libusb_device_handle

    # get handle from file descriptor
    handle = libusb1._libusb_device_handle()
    libusb1._check(lib.libusb_wrap_sys_device(ctx, fd, libusb1.byref(handle)))

    # get device (id?) from handle
    devid = lib.libusb_get_device(handle)

    # device: devid + handle wrapper
    class DummyDevice:
        def __init__(self, devid, handle):
            self.devid = devid
            self.handle = handle
    dev = DummyDevice(devid, handle)

    # create pyusb device
    device = usb.core.Device(dev, backend)
    device._ctx.handle = dev

    # device.set_configuration()
    return device
```

I only tested it on the most current Android (10) and Termux version ...
Usage examples are in my Repo.

**Can we implement this in your library?**

Enumerating devices would maybe work with subprocesses but working with a specific device is only possible in a process(?) callback and does not seem feasable. I guess ..?

jonasmalacofilho:
This seems reasonable.

[...]

PR #287, opened 2019-12-29
Termux (Android support), usb device from sys device (file descriptor), Issue #285

Querela:
See Issue #285 .

Not sure how to implement tests for this.
Resources are released automatically from system by Android. On *nix, too?

### DIFF
--- a/tools/termux_device_info.py
+++ b/tools/termux_device_info.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+"""\
+Run in Termux on Android. Prints device infos for selected device.
+
+See: https://wiki.termux.com/wiki/Termux-usb
+
+Call script in Termux like this:
+    termux-usb -r -e ./tools/termux_device_info.py /dev/bus/usb/001/002
+
+With /dev/bus/usb/001/002 being the correct device path as returned
+by `termux-usb -l`.
+
+If using a virtual environment then call the python script from
+inside a bash script wrapper:
+
+    #!/bin/bash
+    source /path/to/venv/bin/activate
+    python script.py "$@"
+    deactivate
+
+"""
+
+from __future__ import print_function
+
+import usb.core
+
+
+def main(fd):
+    device = usb.core.device_from_fd(fd)
+    print(device)
+
+
+if __name__ == "__main__":
+    # grab fd number from args
+    #   (from termux wrapper)
+    import sys
+
+    fd = int(sys.argv[1])
+    main(fd)
--- a/usb/backend/libusb1.py
+++ b/usb/backend/libusb1.py
@@ -647,6 +647,16 @@ class _DeviceHandle(object):
         self.devid = dev.devid
         _check(_lib.libusb_open(self.devid, byref(self.handle)))
 
+class _FileDescriptorDeviceHandle(object):
+    def __init__(self, fd, ctx):
+        # get handle from file descriptor
+        self.handle = _libusb_device_handle()
+        _check(_lib.libusb_wrap_sys_device(ctx, fd, byref(self.handle)))
+
+        # get device (id?) from handle
+        self.devid = _lib.libusb_get_device(self.handle)
+
+
 class _IsoTransferHandler(_objfinalizer.AutoFinalizedObject):
     def __init__(self, dev_handle, ep, buff, timeout):
         address, length = buff.buffer_info()
@@ -802,6 +812,10 @@ class _LibUSB(usb.backend.IBackend):
     @methodtrace(_logger)
     def close_device(self, dev_handle):
         self.lib.libusb_close(dev_handle.handle)
+
+    @methodtrace(_logger)
+    def get_device_from_fd(self, fd):
+        return _FileDescriptorDeviceHandle(fd, self.ctx)
 
     @methodtrace(_logger)
     def set_configuration(self, dev_handle, config_value):
--- a/usb/backend/libusb1.py
+++ b/usb/backend/libusb1.py
@@ -310,11 +310,24 @@ def _setup_prototypes(lib):
     # void libusb_unref_device(libusb_device *dev)
     lib.libusb_unref_device.argtypes = [c_void_p]
 
+    # int libusb_wrap_sys_device(libusb_context *ctx,
+    #                            intptr_t sys_dev,
+    #                            libusb_device_handle **dev_handle)
+    lib.libusb_wrap_sys_device.argtypes = [
+            c_void_p,
+            c_int,
+            POINTER(_libusb_device_handle)
+        ]
+
     # int libusb_open(libusb_device *dev, libusb_device_handle **handle)
     lib.libusb_open.argtypes = [c_void_p, POINTER(_libusb_device_handle)]
 
     # void libusb_close(libusb_device_handle *dev_handle)
     lib.libusb_close.argtypes = [_libusb_device_handle]
+
+    # libusb_device * libusb_get_device(libusb_device_handle *devh)
+    lib.libusb_get_device.argtypes = [c_void_p]
+    lib.libusb_get_device.restype = _libusb_device_handle
 
     # int libusb_set_configuration(libusb_device_handle *dev,
     #                              int configuration)
--- a/usb/core.py
+++ b/usb/core.py
@@ -1289,3 +1289,18 @@ def show_devices(verbose=False, **kwargs):
             strings += "%s\n\n" % str(device)
 
     return _DescriptorInfo(strings)
+
+def device_from_fd(fd):
+    import usb.backend.libusb1 as libusb1
+
+    backend = libusb1.get_backend()
+    if not backend:
+        raise NoBackendError('No backend available')
+
+    dev = backend.get_device_from_fd(fd)
+
+    # create device
+    device = Device(dev, backend)
+    device._ctx.handle = dev
+
+    return device


Querela:
Hmm, would there be a need for error handling for possible older libusb1 versions? (Where the interface may not exist?)

tools/termux_device_info.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+"""\
+Run in Termux on Android. Prints device infos for selected device.
+
+See: https://wiki.termux.com/wiki/Termux-usb
+
+Call script in Termux like this:
+    termux-usb -r -e ./tools/termux_device_info.py /dev/bus/usb/001/002
```
normanr, line 8: What about making this a generic device_info.py script that supports both -d [vendor]:[product] and --usb-fd [fd] modes. That would make it generally useful and would serve as an example of how other apps could provide similar command line arguments. [argparse](https://docs.python.org/library/argparse.html) would make this pretty easy to do.

Querela: Will _unfortuately_ not work as the device path is part of the termux-api command. My script is called as callback with the open file descriptor as argument. After my script finishes, the termux-api will close the file descriptor.
_You can imagine this like python `with` or JavaScript promises/callbacks._
So my script will have nothing to do with the actual device path.
You _could_ make the script a wrapper around the termux-api (`termux-usb`) command and do some magic with argument handling to also work as the callback but not sure if this is worth the effort. (in my use case)

normanr: Right, it was more mean to make the script generally useful outside of the termux context too.

The usage would look like:
```
device_info.py [-d [vendor]:[product]] [--usb-fd [fd]]
```

so then it could be used on termux (non-root) like this:
```
termux-usb -e 'device_info.py --usb-fd' -r /dev/bus/usb/001/002
# which runs: device_info.py --usb-fd 7
```

and with termux (root), or on other systems as:
```
device_info.py -d 1d6b:0002
device_info.py -D /dev/bus/usb/001/001
device_info.py -s 1:1
```

It _is_ actually possible to support `-D` under termux (non-root): see the `open_with_termux_usb function` in https://gist.github.com/normanr/9816a5db24dda0ce87a17da342a9684c.

I'd actually recommend that the new API should rather be more like the builtin `open` function that takes a path-like object or an integer file descriptor (like the gist's `open` function). The function doesn't even need to differentiate the type of the passed argument and can just call `open(file, mode='r+b', buffering=0)` and use `f.fileno()` to get the file descriptor to create the device handle (note that the file object _must_ be referenced so that's not closed early).

termux-packages could then provide a version of pyusb appropriately patched to convert the path-like object into a file descriptor using `open_with_termux_usb`.

Typical usage would look like:
```
  dev = usb.core.open('/dev/bus/usb/001/001')
```
or if you have a file descriptor:
```
  dev = usb.core.open(7)
```

(termux/termux-api#349 would basically do this (and more) inside of libusb, but I don't think it'll be fixed any time soon).

Querela: Wow. That's interesting and looks promising.
For me, it's somewhat in the past but I remember looking at the socket rec/send functions in python but I was not familiar with them, so I was hesitating. (I was also often looking at other code in C that used this successfully but reimplementing in Python took much effort ...)
I was also considering writing my own Java Termux extensions. (the code is on GitHub so it might have worked but that's another rabbithole)

Currently, I'm still hoping for updates in termux-api and/or pyusb to make my life easier ...

I will have to test your code :-D

Querela: If it takes too long, maybe an alternative solution would be a separate python package. It should be possible to override (extend?) another module I think. (I saw it for example with `faiss` and `faiss-gpu` and the installation order matters but might be solved with dependencies in `setup.py`.)
So Termux users could install this termux-callback 'hack'(?).
But generally the ability to get the device from the `fd` should be useful on any plattform.

normanr: The improvements to the socket library are new in Python 3.3 (before that you had to use libraries like [PyXAPI](https://github.com/metricube/PyXAPI) or [python-passfd](https://github.com/NightTsarina/python-passfd)). My code currently uses `device_from_fd` from this pull request (and ideally it would just be incorporated into this pull request).

I think it would go good to support this _without_ having any termux specifics in the main pyusb branch. pyusb could read the command line for an "external open function" from an environment variable (`PYUSB_OPEN`?). The command would be passed two additional arguments: 1) the command to receive the file descriptor, and 2) the name of the file that pyusb wants to open. Then termux could just set `PYUSB_OPEN='termux-usb -r -e'`. (I've updated the gist to do it this way).

normanr: btw, I think that `device_from_path` and the supporting functions shouldn't go into `usb.core`, but rather a new module instead. I can't think of any good names though: `usb.cli`, `usb.helper`, `usb.path`?

Typical usage could just look like this:
```python
dev = usb.helper.device_from_path('/dev/bus/usb/001/001')
```
and then the `usb.core.device_from_fd` API would become a lower-level API that `device_from_path` uses.

usb/core.py
```diff
@@ -1289,3 +1289,18 @@ def show_devices(verbose=False, **kwargs):
...
+
+    backend = libusb1.get_backend()
+    if not backend:
+        raise NoBackendError('No backend available')
+
+    dev = backend.get_device_from_fd(fd)
+
+    # create device
+    device = Device(dev, backend)
+    device._ctx.handle = dev
```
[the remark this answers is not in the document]
Querela: I find it ugly, too, because it uses the private members. But it is necessary to work correctly (not sure why because I tested it long ago).
I simply tried to replicate the same setup steps but also wanted to correctly separate what has to happen where, the interfaces, the wrapper code etc.
As for the `_DeviceHandle`, I'm not sure. Again, too long ago but I might have missed it.
And I wanted it self-contained because as I wrote this, this was mostly intended for Termux. Other OS might not have to go the way of using the `fd` and can just use the path?

usb/backend/libusb1.py
```diff
@@ -634,6 +647,16 @@ def __init__(self, dev):
         self.devid = dev.devid
         _check(_lib.libusb_open(self.devid, byref(self.handle)))
 
+class _FileDescriptorDeviceHandle(object):
+    def __init__(self, fd, ctx):
```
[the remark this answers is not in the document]
Querela: I find it rather interesting and like having others for feedback. But I'm short on time, so changes will not happen for at least two weeks, maybe more. And you might not be able to change code or suggest actual changes?

As you are deep into the whole material and I was thinking of renaming my pull branch because I started with `master` ... maybe you just open another Pull request and reference mine, while I will close but keep the history?

https://stackoverflow.com/a/48480353/9360161

tools/termux_device_info.py
Querela: Hm. Or would we want both `fd` and `path` in the same package, like `cli`.

`usb.files` might also be valid but would be confusing?
`helper` and the existing `util` seem somewhat similar. 
Do we have more functions for `path`?
`cli` is often a module (in other packages) that contains functions that work as entrypoints, like `Click`, `argparse` etc. But not sure if this is convention or I just stumbled over it so much?

Querela: Termux-packages might work. I'm just not sure if a termux-packages patch might be too hidden? It also might not work with virtual envy depending on how it is made.

I think it would be simpler and more verbose to use a python package that just extends pyusb with the wanted functionality? With version pinning of pyusb there should be no way that something breaks unexpectedly in updates.

usb/backend/libusb1.py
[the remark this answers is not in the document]
normanr: It looks like the primary location for getting feedback is the [mailing list](https://sourceforge.net/projects/pyusb/lists/pyusb-users). It doesn't look it has any mention of termux in the archives..

Nothing is urgent. For now I have something working with just using a local branch with some changes. I don't think I'll have any time in the next week or so to poke at it again.

tools/termux_device_info.py
normanr: I'm not sure if the `fd` and `path` functions need to be in the same package, mostly because `wrap_fd` is a pretty core function, and there's no other place to implement it but in `usb.core`. 

Good point about virtual envs, that is a good idea to not try make a termux specific package. Also termux-packages prefers to not package packages available via language specific ecosystems (eg: [my request to package more-itertools](https://github.com/termux/termux-packages/issues/5756)).

I'd like to avoid having to alter existing code to get it to work with termux - which is why I thought that adding _generic_ support to pyusb for open helpers configured via the environment variable might be acceptable. I also briefly tried proot to bind /dev/bus/usb/ to a temporary directory with a symlink to /proc/self/fd/7, but Android's selinux policy didn't like that.

If adding external opener (or just `path`) support is too hacky for pyusb (it probably is), then just adding the plain `fd` function would be extremely helpful. It would be the responsibility of the main script (or other non-pyusb library) to handle `fd` or `path` arguments on the command line (which I think is why I suggested making the device_info.py script generic, so that it could show how to support `-d vendor:product`, `-D device`, and `--usb-fd fd` as command line arguments with the library changes as were initially proposed).

normanr: example device_info.py: https://gist.github.com/normanr/93c89ab4f1ea9b6c450dae2528eeef25

Querela: Hmm. An `external` or `opener` module?
Too hacky probably not as the whole PyUSB is a wrapper around external USB libraries. It just has to be clear.

Good point about `fd`. But I was thinking more about your snippets with recv/send for `fd`. Or your example `usb.core.open` function that can be in the extra module.

usb/backend/libusb1.py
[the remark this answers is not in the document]
Querela: Ok.
I'm in the same situation as I also have a collection of scripts to patch/extend it for my use-cases...

tools/termux_device_info.py
normanr: Actually I have a better suggestion for the environment variable: instead of it defining a command to use as an external opener it would define/redefine the function used to open the device (similar to how `PYTHONBREAKPOINT` changes the behavior of breakpoint() in [PEP 553](https://www.python.org/dev/peps/pep-0553/)).

That way usb.core could have a very simple `open_device` function that works for most use cases and it can very easily be overridden to support environments like termux. The function would take a `str` (the device path) and return `usb.core.Device`. The default implementation would just `return wrap_fd(os.open(device_path, is.O_RDRW))`.

Then the termux support can exist in another package that is not a core part of pyusb and is easy to install and enable without having to change existing scripts (although they do need to use the `open_device` function). I wonder if there's a need to support existing scripts that only use `find`.

At that point it might be a smaller change to pyusb to make the list of backends configurable with an environment variable instead. Then termux support could be added as a pyusb backend and it would support `find` and all existing scripts that use pyusb without any changes required to use a new `open_device` function. I like this option the most because it's the lease invasive and most powerful (and basically what https://github.com/termux/termux-api/issues/349 would do, except implemented as a pyusb backend instead of inside libusb1)

mcuee:
As mentioned in https://github.com/pyusb/pyusb/issues/285, I am not able to figure out a way to run the demo script under Android (rooted or unrooted) with real HW.

Plain libusb/pyusb work well with the Android 7.1 box with root or without root. Of course I can not do much with root because of permission issues.

mcuee:
All in all, it seems to me for this pull request to work, "termux-usb -l" should work and it seems to me that means the right permission and existence of "/dev/bus/usb" which does not seem to be the case for my Android devices (non rooted).

And this bridge seems to rely on a libusb fork.
https://github.com/normanr/termux-libusb-bridge/blob/main/termux-libusb-bridge.py

normanr:
The bridge requires this libusb fork: https://github.com/green-green-avk/libusb/tree/v1.0.23-android-libusbmanager (from: https://github.com/termux/termux-api/issues/349)

If you can't get "termux-usb -l" to show something on a non-rooted device, then it doesn't what other code you use, it won't work. Once you can get "termux-usb -l" to show something then termux_device_info.py should work with the example command line (replace the usb device path with the path shown by "termux-usb -l"). Note that those paths will not be visible from the termux shell, they only make sense for the server that the termux-usb tool connects to.

You could also try AnotherTerm, usb should "just work" there as far as I understand, ref: https://green-green-avk.github.io/AnotherTerm-docs/installing-libusb-for-nonrooted-android.html

mcuee:
> You could also try AnotherTerm, usb should "just work" there as far as I understand, ref: https://green-green-avk.github.io/AnotherTerm-docs/installing-libusb-for-nonrooted-android.html

Somehow I could not get this AnotherTerm to work. Guess I need to try a bit more. It seems to be pretty complicated.

mcuee:
> If you can't get "termux-usb -l" to show something on a non-rooted device, then it doesn't what other code you use, it won't work. 

That seems to be the key here. I am not so sure how to get that work in the first place...

mcuee:
@normanr 
From https://github.com/termux/termux-api/issues/349
> If I remember correctly I compiled it within Termux (using instructions in `android/README`). I didn't install it, I just added `export LD_LIBRARY_PATH=$HOME/src/libusb/libusb/.libs/` to `.profile`. I also created a service (under $PREFIX/usr/var/services) to start the bridge automatically.

So there are a few steps here, right?
1) first step is to compile the following libusb fork under Termux
https://github.com/green-green-avk/libusb/tree/v1.0.23-android-libusbmanager

2) then the second step is to run the bridge here
https://github.com/normanr/termux-libusb-bridge
3) optional: create a service to run the bridge automatically

mcuee:
I think this PR is in the right direction but not ready yet.

libusb current ways to work under Android (without root).
https://github.com/libusb/libusb/tree/master/android
https://github.com/libusb/libusb/tree/master/android/examples

There is also an on-going PR here.
* https://github.com/libusb/libusb/pull/1164

mcuee:
Relevant discussions from termux
* https://github.com/termux/termux-api/issues/349
* https://github.com/termux/termux-packages/issues/5831

The OP's repo:
https://github.com/Querela/termux-usb-python

libusb PR
* https://github.com/libusb/libusb/pull/874 (closed, replace by below one)
* https://github.com/libusb/libusb/pull/1164

PR #297, opened 2020-02-23
replace non-ascii character \xe2

guyingzhao:
replace non-ascii characters to avoid SyntaxError


guyingzhao:
close this since it's already fixed

jonasmalacofilho:
> close this since it's already fixed

What do you mean by "already fixed"?

Issue #63, opened 2014-09-23
Implement asynchronous transfer functions


walac:
Implement asynchronous versions of the transfers functions.

jrast:
This is a very old issue, any updates on the state of this feature?

Armavica:
I would like to help make this happen. Where do we start?

SimplicityGuy:
How familiar are you with pyusb internals? If you are, please create a branch with these changes. Myself, along with others, will review the changes.

Armavica:
Ok. Just to check my understanding of what exactly we want to do, the end goal of this would be to implement the [asyncio transfers/protocols](https://docs.python.org/3/library/asyncio-protocol.html) for USB? For this, I need to

1. Write the FFI allowing to call the async libusb function from Python;
2. Connect this to asyncio.

Is that correct?

mrnuke:
Hi, I'm from the future, in the year 2020. I believe I have a compelling use case for this. I have a device here (Agilent 82347A) that needs transfers submitted in a very specific way:

1. Submit a BULK IN transfer
2. Send a BULK OUT transfer with a commant to read data
3. Receive data via a BULK IN endpoint

With the available API, the best we can do is:
1. Submit and finish BULK OUT transfer
2. Submit and receive BULK IN transfer

What can happen in this case is, the device processes the command in step (1), sees no active IN request, and discards the data. When the IN request arrives, there is no data to send back, and the transfer times out.

jonasmalacofilho:
Hi @mrnuke,

Do you think you can help me with this?

You could start by adding the [additonal libusb async functions](http://libusb.sourceforge.net/api-1.0/group__libusb__asyncio.html) to the [`libusb1`](https://github.com/pyusb/pyusb/blob/master/usb/backend/libusb1.py) backend.

---

I still don't have a complete picture of how that should be exposed on the higher level and backend-agnostic PyUSB layer.  Also, @Armavica mentioned implementing Python's [asyncio](https://docs.python.org/3/library/asyncio.html), which might indeed be a good idea.

Anyway, that can wait until we have the C FFI ready, and you could probably use it directly in pygpib until we figure out and implement the higher-level APIs.

mrnuke:
I was just looking into calling libusb_submit_transfer on top of pyusb, albeit in a hackier manner.

mrnuke:
HI @jonasmalacofilho ,

I tried to implement libusb_fill_bulk_transfer() here 
[...]

PR #303, opened 2020-04-12
[WIP] Begin implementation of asynchronous transfers

mrnuke:
The purpose of these patches is to allow asynchronous USB transfers. Keep in mind that I am not claiming this should be the final form. I'm not an asyncio expert, and I am open for suggestions on how to improve the implementation. Note that async transfers are only imlemented for the libusb1 backend.
I am striving to achieve an API that looks like the following example:

	async def do_usb_stuff():
	    dev = usb.core.find(...)
	
	    transfer1 = dev.submit_read(...)
	    transfer2 = dev.submit_write(...)
	
	    await transfer2.result()
	    await transfer1.result()

Note that this is symmetrical to the sync functions, in that the transfer type does not have to be specified. It is derived from the endpoint.

await transfer.result() is similar to an asyncio.Future, although the implementation is not exactly an awaitable future. This is an implementation detail, and it could be changed if needed.

### DIFF
--- a/usb/backend/libusb1.py
+++ b/usb/backend/libusb1.py
@@ -512,6 +512,30 @@ def _setup_prototypes(lib):
     lib.libusb_get_max_iso_packet_size.argtypes = [c_void_p,
                                                    c_ubyte]
 
+    # void libusb_fill_bulk_transfer(
+    #               libusb_transfer* transfer
+    #               libusb_device_handle* dev_handle
+    #               char endpoint
+    #               char* buffer
+    #               int length
+    #               libusb_transfer_cb_fn callback
+    #               void* user_data
+    #               uint timeout
+    #           );
+    def libusb_fill_bulk_transfer(_libusb_transfer_p, dev_handle, endpoint,
+                                  buffer, length, callback, user_data, timeout):
+        r"""This function is extremely dangerous, so we must deal with it."""
+        transfer = _libusb_transfer_p.contents
+        transfer.dev_handle = dev_handle
+        transfer.endpoint = endpoint
+        transfer.type = _LIBUSB_TRANSFER_TYPE_BULK
+        transfer.timeout = timeout
+        transfer.buffer = cast(buffer, c_void_p)
+        transfer.length = length
+        transfer.user_data = user_data
+        transfer.callback = callback
+    lib.libusb_fill_bulk_transfer = libusb_fill_bulk_transfer
+
     # void libusb_fill_iso_transfer(
     #               struct libusb_transfer* transfer,
     #               libusb_device_handle*  dev_handle,
--- a/usb/backend/libusb1.py
+++ b/usb/backend/libusb1.py
@@ -536,6 +536,30 @@ def _setup_prototypes(lib):
         transfer.callback = callback
     lib.libusb_fill_bulk_transfer = libusb_fill_bulk_transfer
 
+    # void libusb_fill_interrupt_transfer(
+    #               struct libusb_transfer* transfer,
+    #               libusb_device_handle* dev_handle,
+    #               unsigned char endpoint,
+    #               unsigned char* buffer,
+    #               int length,
+    #               libusb_transfer_cb_fn callback,
+    #               void* user_data,
+    #               unsigned int timeout
+    #           );
+    def libusb_fill_interrupt_transfer(_libusb_transfer_p, dev_handle, endpoint,
+                                  buffer, length, callback, user_data, timeout):
+        r"""This function is extremely dangerous, so we must deal with it."""
+        transfer = _libusb_transfer_p.contents
+        transfer.dev_handle = dev_handle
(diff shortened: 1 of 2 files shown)


mcuee:
https://github.com/vpelletier/python-libusb1
Not so sure if the implementation in python-libusb1 is of some reference value here.

mcuee:
This is a WIP to address #63.

usb/backend/libusb1.py
```diff
@@ -651,6 +700,107 @@ def __init__(self, dev):
...
+        return self.fut.result()
+
+    async def _usb_event_loop(self):
+        """ Helper function to process USB events and complete futures
+
+        TODO: This processes all events in the current libusb context, self._ctx
+        It seems benign, since transfers that complete earlier will have been
+        marked as donem and not run libusb_handle_events(). However, is this a
+        correct way to handle events?
+        TODO 2: Is libusb_handle_events() really non-blocking?
```
WhiteApfel, line 745: Nope 😕 I tried to test on my receipt printer:
```python
from escpos.printer import Usb

p = Usb(0x2730, 0x2002, None, 0, 0x81, 0x02)
p.set('left')
p.text(f"{'ПУТИН ХУЙЛО ' * 300}")
p.qr("All the sings she said", size=16, model=2)
p.cut(feed=False)
````

Changed the function to show execution time:
```python
        import time
        while not self._fut.done():
            st = time.time()
            _check(_lib.libusb_handle_events(self._ctx))
            print(round(time.time()-st, 5))
            await asyncio.sleep(0.0001)
```

```
6e-05
4e-05
3e-05
3e-05
3e-05
3e-05
3e-05
5e-05
3e-05
1e-05
# start of printing qr
0.01008
# end of printing qr
1e-05
# start of printing "путин хуйло"
6.02684
# end of printing "путин хуйло"
4e-05
3e-05
3e-05

```

mcuee:
Not sure if anyone is working on this or not. I tend to think this is one major missing feature (along with hotplug).

grvstick:
I don't know a lot about how `ctypes` work but I think the transfer class should be looking like this, according to libusb document. It takes advantage of file descriptors for platforms that can use it. If not, which is Windows case, it falls back to using `run_in_executor`, which is using threads. `libusb_handle_events` seems indeed to be a `blocking` function but sits idle waiting for events, so using threads could make sense here.

```python
class _AsyncTransfer():
    """ Generic API for asynchronous USB transfers"""
    def __init__(self, dev_handle, ep, buf, iso_packet_count, timeout):
        self._transfer = _lib.libusb_alloc_transfer(iso_packet_count)
        self._ctx = None
        self._buf = buf
        self._event_term_event = threading.Event()
        self._fds = []
        
        if "windows" in platform.platform().lower():
            self._threaded = True
        else:
            self._threaded = False

    def __del__(self):
        if not self._fut.done():
            _lib.libusb_cancel_transfer(self._transfer)

        _lib.libusb_free_transfer(self._transfer)

    def submit(self, ctx = None):
        """ Submit the USB transfer, but don't wait for completion

        TODO: Is it better to create the Future in __init__ and reuse it?
        TODO 2: Exception when libusb_submit_transfer() fails()
        """
        self._fut = asyncio.Future()
        self._ctx = ctx
        self._event_term_event.clear()
        _check(_lib.libusb_submit_transfer(self._transfer))

        return self

    def _add_usb_event_handler(self):
        """ Helper function to process USB events and complete futures
        """

        self._fds = _lib.libusb_get_pollfds(self._ctx)
        loop = asyncio.get_event_loop()
        for fd in self._fds:
            loop.add_reader(fd, self._fd_reader_callback)
        
    def _usb_event_loop_threaded(self):
        """
            This should be run in separate thread. The way it is used in this class is not for "hotplug" case.
        """
        while not self._event_term_event.is_set():
            _check(_lib.libusb_handle_events(self._ctx))
            # add 0 sec sleep here so the threads can find their optimized path
            time.sleep(0)

    async def result(self):
        """ 
        Asynchronous result of USB transfer. Should be 'await'ed
        TODO: handle timeout cases
        """
        if self._threaded is True:
            loop = asyncio.get_event_loop()
            await loop.run_in_executor(None, self._usb_event_loop_threaded)

        return await self._fut.result()

    def _fd_reader_callback(self):
        _lib.libusb_handle_events_timeout_completed(self._ctx, POINTER(timeval(0, 0)), None)

    def _callback(self, transfer):
        status = int(transfer.contents.status)
        if status != LIBUSB_TRANSFER_COMPLETED:
            result =  usb.USBError(_str_transfer_error[status],
                                  status, _transfer_errno[status])
        else:
            result = self._buf[:transfer.contents.actual_length]

        # TODO: this next line should be tested if does not violate thread safety.
        self._fut.set_result(result)
        if self._threaded is True:
            self._event_term_event.set()

```

I would like to test this code but need a test device and look deeper into `ctypes`.

PR #306, opened 2020-05-17
a bit spell fix in docstring.

kuma35:
fix interrut to interrupt


jonasmalacofilho:
Thank you!

P.S. I saw your message on the mailing list, but this way (a pull request) is better.  I also have issues replying to that list.

Issue #298, opened 2020-02-25
Version number is out-of-sync with Git code


eblot:
Hi,

I'm the maintainer of PyFtdi. I had some hard time figuring out why some reported issues with PyFtdi depicting PyUSB traces did not match at all my own installation of PyUSB. Even after reinstalling from scratch, clearing out pip caches, etc, there were obvious discrepancies between packaged 1.0.2 version available from PyPi and the traces I was reading.

PyUSB has resumed development and that's really great.

However the version number has not been updated and is still defined as `1.0.2`. Current delta with `1.0.2` is:  `30 files changed, 775 insertions(+), 630 deletions(-)`. This is really huge.

It would be nice to add a -dev [or any common suffix](https://www.python.org/dev/peps/pep-0440/#pre-releases) to the `__init__.py` file so that there is no confusion with the active development version and the packaged version on PyPi. I guess PyFtdi will not be the only package which depends on PyUSB and will encounter some maintenance issues to deal with, and confusion is likely to occur.

Note: I used to push PyUSB for the original author (Wander) to [PyPi](https://pypi.org/project/pyusb/#history), so if you do not have access yet to PyPi for PyUSB, please contact me.

Thanks.

jonasmalacofilho:
Hi again,

I'll consider doing this after the 1.1.0 release.

But I haven't seen this work particularly well in my own stuff: it still is a static version number shared by many commits.

> Additional note: it seems you have not installed PyUSB from an official package but a development version.
—_(from the downstream issue)_

Regarding stable vs. development releases, the most common cases (pip install git or --editable) can be handled by asking users to provide the output of `pip list`.

Nonetheless, I'll give it some (thorough) thought.

eblot:
Well, I agree it is not perfect, at least it gives an instant clue from the traceback *e.g.* `site-packages\pyusb-1.0.2-dev-py3.6.egg` vs. `site-packages\pyusb-1.0.2-py3.6.egg`., i.e. is it a release or any development version.

To avoid static version, you may add the git distance from the last official tag:

    $ git log -1
    commit cf88dfcd3b38afd00ec9fc7229213c8e8aa42118 (HEAD -> master, origin/master, origin/HEAD)
    Author: root <572488191@qq.com>
    Date:   Sun Feb 23 20:06:06 2020 +0800
    
        Remove non-ASCII dashes from copyright notices (#297)
    
        In Python 2 source files are treated as encoded in ASCII.
[...]

PR #307, opened 2020-06-09
Use setuptools_scm for versioning

ap--:
Automatically generates the `usb.__version__` string on installation.
Would close pyusb/pyusb#298.

**This PR adds `setuptools_scm` to automatically generate version strings on install.**

There's still a few things to be considered before this should be merged:
- what are the target python versions and OSes? (Let's create a list here and I can work my way through all of them to see if packaging won't fail due to old setuptools versions)
- I set the setuptools_scm `version_scheme` to **post-release** which generates version strings like `1.0.2.post42+SHORTGITHASH` (42 commits after tag '1.0.2') instead of the default which would increase the micro version and create `1.0.3.dev42+SHORTGITHASH`, but that's just personal preference I guess. (see [here](https://github.com/pypa/setuptools_scm#version-number-construction))
- I set the fallback version to `0.0.0` in case something went wrong with loading `usb/_version.py`. The only scenario I can come up with in which this should happen is, if someone clones the repo and imports pyusb from the repository root without installing (this is the default when running the tests manually right now).
- the `usb.version_info` tuple is generated from the version string and supports the tag format that's already used in the repository. So for example:
  ```
  # git tag: v1.0.0
  __version__ = "1.0.0"
  version_info = (1, 0, 0)

  # git tag: v1.0.0rc1
  __version__ = "1.0.0rc1"
  version_info = (1, 0, 0, 'rc1')

  # one commit newer than tag v1.0.0
  __version__ = "1.0.0.post1+ace0f2e4"
  version_info = (1, 0, 0, 'post1+ace0f2e4')

  # one commit newer than tag v1.0.0 AND repository has uncommited changes
  __version__ = "1.0.0.post1+ace0f2e4.d20200609"
  version_info = (1, 0, 0, 'post1+ace0f2e4.d20200609')
  ```

I also tried to introduce as few changes as possible.
Happy to get some feedback on the PR and to discuss the mentioned points above.

Cheers,
Andreas :smiley:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ MANIFEST
 dist
 .tox
 *.egg-info
+usb/_version.py
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools>=30.3.0", "wheel", "setuptools_scm"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [metadata]
 description-file = README.rst
+
+[options]
+setup_requires = setuptools_scm
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,12 @@
 
 
 from setuptools import setup
+from distutils.version import LooseVersion
+from setuptools import __version__ as setuptools_version
 
+setuptools_scm = 'setuptools_scm'
+if LooseVersion(setuptools_version).version[0] < 12:
+    setuptools_scm += '<2.0'
 
 setup(
     name='pyusb',
@@ -39,6 +44,7 @@ setup(
         "version_scheme": "post-release",
         "write_to": "usb/_version.py",
     },
+    setup_requires=setuptools_scm,
     description='Python USB access module',
     author='Robert Wlodarczyk',
     author_email='robert@simplicityguy.com',
--- a/setup.py
+++ b/setup.py
@@ -32,12 +32,13 @@
 
 from setuptools import setup
 
-import usb
-
 
 setup(
     name='pyusb',
-    version=usb.__version__,
+    use_scm_version={
+        "version_scheme": "post-release",
+        "write_to": "usb/_version.py",
+    },
     description='Python USB access module',
     author='Robert Wlodarczyk',
     author_email='robert@simplicityguy.com',
--- a/usb/__init__.py
+++ b/usb/__init__.py
@@ -48,8 +48,19 @@ import os
 __author__ = 'Wander Lairson Costa'
 
 # Use Semantic Versioning, http://semver.org/
-version_info = (1, 0, 2)
-__version__ = '%d.%d.%d' % version_info
+try:
+    from usb._version import version as __version__
+except ImportError:
+    __version__ = '0.0.0'
+
+def _get_version_info(version):
+    import re
+    m = re.match(r'(\d+)\.(\d+)\.(\d+)[.-]?(\D.*)?', version)
+    return tuple(
+        int(v) if v.isdigit() else v for v in m.groups() if v
+    )
+
+version_info = _get_version_info(__version__)
 
 __all__ = ['legacy', 'control', 'core', 'backend', 'util', 'libloader']
 
--- a/usb/__init__.py
+++ b/usb/__init__.py
@@ -53,14 +53,14 @@ try:
 except ImportError:
     __version__ = '0.0.0'
 
-def _get_version_info(version):
+def _get_extended_version_info(version):
     import re
-    m = re.match(r'(\d+)\.(\d+)\.(\d+)[.-]?(\D.*)?', version)
-    return tuple(
-        int(v) if v.isdigit() else v for v in m.groups() if v
-    )
+    m = re.match(r'(\d+)\.(\d+)\.(\d+)[.-]?(.*|\b)', version)
+    major, minor, patch, suffix = m.groups()
+    return int(major), int(minor), int(patch), suffix
 
-version_info = _get_version_info(__version__)
+extended_version_info = _get_extended_version_info(__version__)
+version_info = extended_version_info[:3]
 
 __all__ = ['legacy', 'control', 'core', 'backend', 'util', 'libloader']
 
--- a/usb/__init__.py
+++ b/usb/__init__.py
@@ -55,7 +55,7 @@ except ImportError:
 
 def _get_extended_version_info(version):
     import re
-    m = re.match(r'(\d+)\.(\d+)\.(\d+)[.-]?(.*|\b)', version)
+    m = re.match(r'(\d+)\.(\d+)\.(\d+)[.-]?(.*)', version)
     major, minor, patch, suffix = m.groups()
     return int(major), int(minor), int(patch), suffix
 


jonasmalacofilho:
> what are the target python versions and OSes?  [...]

I think you already got the Python versions right in #308.  More specifically:

- Target: Python >=3.6
- Support (at least until the next PyUSB release/the end of the year): 2.7 and 3.5
- Ignore (but don't _intentionally_ break until the next PyUSB release/the end of the year): <=2.6, <=3.4

There are no specific OS targets; ideally in any OS where Python and libusb-1.0 are supported PyUSB should also work.

Besides what you're already testing with #308, here we probably also want to make sure that, at least in theory, things will work smoothly on the BSDs and the popular (but sometimes old) Linux distros that haven't yet reached EOL.  For the latter case, all currently supported Ubuntu versions is a reasonable (if incomplete) benchmark.

Another detail is that on Linux distros that may update PyUSB to 1.1.0, it is important that it builds without issues with the setuptools and setuptools_scm versions supplied by the distro.

---

> I set the setuptools_scm `version_scheme` to **post-release** [...]

I also prefer this because it doesn't increment the base version.  A local build from git should not assume that any particular commit since the last tag will be included in the next patch (or minor) release.

These wont be published, so I don't think that ([arguably](https://www.python.org/dev/peps/pep-0440/#id28)) misusing the post-release tags is an issue.

---

> I set the fallback version to 0.0.0 in case something went wrong with loading usb/_version.py

Looks good to me.

---

> the `usb.version_info` tuple is generated from the version string and supports the tag format that's already used in the repository. So for example: [...]

I think it's important for tuples to always have the same number of elements, so it's trivial to destructure them.

Additionally, while there is precedent in the repository for version_info to have a fourth segment, that was removed when 1.0.0 was bumped to sable (d7e6291564a5d5683a8b03fc9a761ad3e3dd70ea).  Also because it's a public field, I think it should be kept with only three segments (the major/minor/patch numbers).

A new `extended_version_info` attribute could be added to hold the fourth segment or, alternatively, to hold a tuple with four segments.

ap--:
Hi @jonasmalacofilho 

I implemented the version_info fix as you suggested.
`usb.version_info` is a (major, minor, patch) tuple and
`usb.extended_version_info` adds the suffix to return a length 4 tuple (major, minor, patch, suffix).
For releases the suffix will be an empty string.

Package maintainers can potentially use this to add their distribution specific build string if they patched something (see: commit message in f71fc4f600c065f4932f20fd1676d5669eca4df0)


I'm currently checking if installing from a source tarball from pypi would cause problems. So far i tested:


**As `python[23]?-usb.*`**

| OS | python | setuptools | setuptools-scm |
|----|--------|------------|----------------|
| - [x] Debian 10 Buster | 3.7.3 | 40.8.0 | 3.2.0 |
| - [x] Debian 9 Stretch | 2.7.13 | 33.1.1 | 1.15.0 |
| - [ ] Debian 8 Jessie | 2.7.9 | 5.5 | - |
| - [x] Ubuntu 20.04 | 3.8.2 | 45.2.0 | 3.4.3 |
| - [x] Ubuntu 19.10 | 3.7.5 | 41.1.0 | 3.3.3 |
| - [x] Ubuntu 18.04 | 2.7.15rc1 | 39.0.1 | 1.15.6 |
| - [x] Ubuntu 16.04 | 3.5.1 | 20.7.0 | 1.10.1 |
| - [ ] Ubuntu 14.04 | 2.7.5 | 3.3 | - |

The debians and ubuntus all install fine. Problematic are jessie (soon EOL) and 14.04 (already EOL) due to horribly outdated setuptools. Interestingly running `python setup.py install` installs fine, but creates a wrong egg version (0.0.0). I'll have to look into this more what the best fallback looks like when setuptools is outdated and setuptools-scm is not installed.


**As `python[23]?-pyusb.*`**

| OS | python | setuptools | setuptools_scm |
|----|--------|------------|----------------|
| - [x] CentOS 8 | 3.6.8 | 39.2.0 | 1.15.7 |
| - [ ] CentOS 7 | 2.7.5 | 0.9.8 | - |
| - [ ] CentOS 6 | 2.6.6 | 0.6.10 | - |
| - [ ] RHEL 8.1  | 3.6.8 | ? | ? |
| - [ ] RHEL 7.8  | 3.6.8 | ? | ? |
| - [ ] RHEL 6.9  | 2.6.6 | 0.6.10 | - |
| - [ ] RHEL 5.11 | 2.4.3 | 0.6c5 | - |
| - [x] Fedora 32 | 3.8.2 | 41.6.0 | 3.3.3 |
| - [x] Fedora 31 | 3.7.4 | 41.x.x | 3.3.3 |
| - [x] Fedora 30 | 3.7.3 | 40.8.0 | 3.2.0 |

Installation works on CentOS 6,7,8 but again as for Debian-Jessie CentOS6 and CentOS7 generate incorrect egg version (0.0.0). The currently supported Fedoras all work. I haven't tested any of the RHELs but I'd assume a similar outcome (I'd guess it'll install on all of them, but egg versions will be 0.0.0 on RHEL6 and RHEL5)


**As `py{27,36}?-usb.*`**

| OS | python | setuptools | setuptools_scm |
|----|--------|------------|----------------|
| - [x] FreeBSD 10.4 | 2.7.15 | 40.4.3 | 1.17.0 |
| - [x] OpenBSD 6.7 | 2.7.18p0 | 41.6.0 | 3.3.3 |  

It installs correctly on OpenBSD 6.7 and FreeBSD 10.4 (which is already unsupported, so it'll work on FreeBSD 11 and 12 too), because of the new setuptools versions and setuptools_scm availability.


I'll see if I can find a solution for the incorrect version on the installs that have very old setuptools versions.

Cheers,
Andreas :smiley:

jonasmalacofilho:
> I'm currently checking if installing from a source tarball from pypi would cause problems.

Thinking a bit more about this, it seems to me that installs from PyPI should be unaffected, because the published sdists will already include a populated `_version` module.¹

So what we really want to know is if building the sdist from git works on each OS, which should cover both distros that build from tags and users wanting to test the latest code.  But based on the results I think this is already what you're testing (please confirm).

> I'll have to look into this more what the best fallback looks like [...]

Assuming what I said above is correct,¹ then one option could be to default to something like `1.0.2+unknown`.

This would be particularly useful in light of the original problem described in #298: end users running the latest PyUSB code from the repository.  And distros with problematic versions of setuptools could always switch to the released sdist, or patch this fallback version.

_¹ We should obviously still check that the problematic distros do indeed correctly install a sdist built with `_version` in some other environment.  My only concern is the lack of a `version` argument in `setup.py`, but I hoping that it isn't used after the sdist has already been made._

ap--:
> So what we really want to know is if building the sdist from git works on each OS, which should cover both distros that build from tags and users wanting to test the latest code. But based on the results I think this is already what you're testing (please confirm).

I managed to fix the issues for the older distros by adding `setup_requires` to the `setup.py` and requesting `setuptools_scm<2.0` in case setuptools is older than version 12, as described [here](https://github.com/pypa/setuptools_scm#interaction-with-enterprise-distributions).

Regarding the test install scenarios:

**git**: https://github.com/ap--/pyusb/actions/runs/149997657
This tests installation from a checkout of the git repository, and works also for Ubuntu 14.04, Debian Jessie and CentOS6.

**sdist**: https://github.com/ap--/pyusb/actions/runs/150028470
This tests installation from an sdist tarball (built on 18.04 with python3). It also works for all of them (note, centos6 fails because the github action for downloading the sdist artifact is incompatible. I have no reason to believe it wouldn't install correctly from the sdist)

> Assuming what I said above is correct,¹ then one option could be to default to something like 1.0.2+unknown.

Do you mean changing the `ImportError` fallback from `'0.0.0' to `'1.0.2+unknown'`?

jonasmalacofilho:
Actually, please disregard that suggestion.  Like you said in the first comment, the only scenario where importing _version fails is when the root is used directly, without installing.

Thinking about the motivation described in #298, I can still see one somewhat big hole: `pip install --editable`.  But, similar to when the root is used directly, there's nothing that can be done.

Thank you so much for figuring all of this out, and for the extensive testing.

jonasmalacofilho:
Well, I mentioned there isn't anything we can do about `pip install --editable`, but that's not entirely true.

First, we could try to detect that this is the case at runtime, and modify `extended_version_info` accordingly.

Another way is to customize the behavior of `python.py develop` and modify the version information there.  I do this [in liquidctl ](https://github.com/jonasmalacofilho/liquidctl/blob/d1b8d2424948c564e218e2f0cf5ffb86f21b1445/setup.py#L57) (although that project certainly needs start using `setuptools_scm` now).

ap--:
> Like you said in the first comment, the only scenario where importing _version fails is when the root is used directly, without installing.

From my experience most "accidental" uses of using the root directly can be prevented by using the 'src layout' for the package (moving the module to a 'src' subfolder), which is still compatible with editable installs.


> Well, I mentioned there isn't anything we can do about pip install --editable, but that's not entirely true. First, we could try to detect that this is the case at runtime, and modify extended_version_info accordingly.

Good point. I think this can be covered by adopting this use case: [importing in setup.py](https://github.com/pypa/setuptools_scm#importing-in-setuppy)

Let me know if you'd like me to try to implement this. I might have a bit of time next weekend :blush: 
I guess it would "just" have to add '+dev' to the version string in case it's an editable install.

Cheers,
Andreas :smiley:

jonasmalacofilho:
> Good point. I think this can be covered by adopting this use case: importing in setup.py

I agree.

> Let me know if you'd like me to try to implement this. I might have a bit of time next weekend blush
> I guess it would "just" have to add '+dev' (or some other indicator) to the version string in case it's an editable install.

Yes, that would be awesome, thanks!

---

We may still be left with a (minor) hole: we're using `write_to` to avoid a runtime call to get the version from metadata so, even after fixing `pip install --editable`, this can happen (I think):

```properties
$ pip install --editable .
$ python -c "import usb; print(usb.__version__)"
1.1.0.post23+0f2e4bc8-editable

$ git checkout v1.1.0
$ python setup.py build                                    # changes _version.py
$ python -c "import usb; print(usb.__version__)"
1.1.0

$ git checkout master
$ python -c "import usb; print(usb.__version__)"
1.1.0                                                      # wrong, should be 1.1.0.post23+0f2e4bc8-editable
```

But I think this minor hole is an acceptable trade-off for not introducing a runtime dependency to `pkg_resources`.

---

As for the specific format, the first idea that comes to my mind also is appending `-dev`, `-editable` or something like that to the version string.  For now I would also keep everything else already in the version string, but maybe post + commit is more likely to induce errors than to give (occasionally) useful information.  What do you think?

PR #308, opened 2020-06-14
run tests via tox and github actions

ap--:
This PR adds a github action workflow to run the tests for python 2.7, 3.5, 3.6, 3.7, 3.8 on windows, linux and macos via tox.

Currently windows-py27 fails due to an ImportError (fix: https://github.com/ap--/pyusb/commit/5ece7fa23597d99abcb62df413ef9d22fcbf3f0d) and all windows tests throw warnings because no backend can be imported.

Since tox runs `python setup.py sdist` to install the package in the test environment, this could be used to verify that `setuptools_scm`'s version is correctly included.

Cheers :smiley: 
Andreas

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,15 +2,6 @@
 envlist = py{27,35,36,37,38}
 requires =
     setuptools >= 30.3.0
-    tox-gh-actions
-
-[gh-actions]
-python =
-    2.7: py27
-    3.5: py35
-    3.6: py36
-    3.7: py37
-    3.8: py38
 
 [testenv]
 changedir=tests
--- a/tox.ini
+++ b/tox.ini
@@ -1,37 +1,17 @@
-# Tox is broken with Python 2.4
-# setuptools is broken with Python 2.5
-# pythonz is broken with Python 3.0
-# Pip 1.5.4 (distributed with Ubuntu 14.04) is broken with Python 3.1
 [tox]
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,37 +1,17 @@
-# Tox is broken with Python 2.4
-# setuptools is broken with Python 2.5
-# pythonz is broken with Python 3.0
-# Pip 1.5.4 (distributed with Ubuntu 14.04) is broken with Python 3.1
 [tox]
-envlist = py26,py27,py32,py33,py34
+envlist = py{27,35,36,37,38}
+requires =
+    setuptools >= 30.3.0
+    tox-gh-actions
```
jonasmalacofilho, line 5: Is this really required?

jonasmalacofilho:
```suggestion
```

jonasmalacofilho:
This is great, thanks a lot!

ap--:
Thanks for the comments :smiley: 
I went over it again and removed the dependency on tox-gh-actions, which only took care of running the correct test env for each job.
So now in the github action I just request tox to only run the specific pyNM env.

Also, I switched to the newer setup-python action, because `v1` relies on the preinstalled python version on the images. So the newer version is more robust, in case they remove 3.5 at some point in the future.

jonasmalacofilho:
Thanks again!

PR #312, opened 2020-07-12
Indicate editable installs in version string

ap--:
This PR adds "-editable" to `usb.__version__` of pyusb installs done via `python setup.py develop` or `pip install -e .`

Cheers,
Andreas :smiley:


jonasmalacofilho:
Thanks again!  And the `pop` fix was ingenuous...

PR #366, opened 2021-05-13
Fixed errno ENODATA use causing unhandled exception.

cederom:
This fix uses syntax described in Python Manual:
https://docs.python.org/3.9/library/errno.html

Signed-off-by: Tomasz 'CeDeROM' CEDRO <tomek@cedro.info>

### DIFF
--- a/usb/backend/libusb0.py
+++ b/usb/backend/libusb0.py
@@ -619,27 +619,47 @@ class _LibUSB(usb.backend.IBackend):
 
     @methodtrace(_logger)
     def is_kernel_driver_active(self, dev_handle, intf):
+        # libusb0's usb_get_driver_np() is platform specific / not portable.
+        # This function checks what kernel module handles dev_handle/intf.
+        # Various implementations may return system specific values.
+
         if not hasattr(_lib, 'usb_get_driver_np'):
             raise NotImplementedError(self.is_kernel_driver_active.__name__)
-        from errno import ENODATA
         buf = usb.util.create_buffer(_USBFS_MAXDRIVERNAME + 1)
         name, length = buf.buffer_info()
         length *= buf.itemsize
-        # based on the implementation of libusb_kernel_driver_active
+
+        # Linux: based on the implementation of libusb_kernel_driver_active
         # (see libusb/os/linux_usbfs.c @@ op_kernel_driver_active):
         # usb_get_driver_np fails with ENODATA when no kernel driver is bound,
         # and if 'usbfs' is bound that means that a userspace program is
         # controlling the device (e.g. using this very library)
+        # FreeBSD: ENODATA is undefied, returns 0 on success, -1 on error.
+
+        if "linux" in sys.platform:
+            try:
+                from errno import ENODATA
+            except:
+                raise NotImplementedError(self.is_kernel_driver_active.__name__)
+
         try:
             _check(_lib.usb_get_driver_np(
                         dev_handle,
                         intf,
                         cast(name, c_char_p),
                         length))
-            return cast(name, c_char_p).value != b'usbfs'
+            if "linux" in sys.platform:
+                # On Linux 'usbfs' indicates userland application.
+                return cast(name, c_char_p).value != b'usbfs'
+            else:
+                # On other platforms just check any module attached.
+                return buf.index(0) != 0
         except USBError as err:
-            if err.backend_error_code == -ENODATA:
-                return False
+            # On Linux system specific ENODATA means no module attached.
+            # On other systems 0 or error does not imply (in)active module.
+            if "linux" in sys.platform:
+                if err.backend_error_code == -errno.ENODATA:
+                    return False
             raise err
 
     @methodtrace(_logger)


hselasky (commented):
Looks good to me.

usb/backend/libusb0.py
```diff
@@ -638,7 +638,7 @@ def is_kernel_driver_active(self, dev_handle, intf):
                         length))
             return cast(name, c_char_p).value != b'usbfs'
         except USBError as err:
-            if err.backend_error_code == -ENODATA:
+            if err.backend_error_code == -errno.errorcode[errno.ENODATA]:
```
jonasmalacofilho, line 641: How can `errno.ENODATA` still be valid if `from errno import ENODATA` fails?

According to the documentation (and to a quick look at the CPython code), that failure implies that ENODATA is not defined for your platform.

Also, on your platform, what do `errno.ENODATA` and `errno.errorcode[errno.ENODATA]` equal to?

cederom: True!!! Good catch! Sorry and thank you :-)

Python 3.7:
```
(venv37zephyr) python
Python 3.7.10 (default, Apr  8 2021, 01:15:14)
[Clang 10.0.1 (git@github.com:llvm/llvm-project.git llvmorg-10.0.1-0-gef32c611a on freebsd12
Type "help", "copyright", "credits" or "license" for more information.
>>> import errno
>>> errno.ENODATA
  File "<stdin>", line 1, in <module>
AttributeError: module 'errno' has no attribute 'ENODATA'
>>> errno.errorcode[errno.ENODATA]
  File "<stdin>", line 1, in <module>
AttributeError: module 'errno' has no attribute 'ENODATA'
```

Python 3.8:
```
hexagon% source venv38bsd/bin/activate
(venv38bsd) hexagon% python
Python 3.8.9 (default, Apr 15 2021, 01:10:27)
[Clang 10.0.1 (git@github.com:llvm/llvm-project.git llvmorg-10.0.1-0-gef32c611a on freebsd12
Type "help", "copyright", "credits" or "license" for more information.
>>> import errno
>>> errno.ENODATA
  File "<stdin>", line 1, in <module>
AttributeError: module 'errno' has no attribute 'ENODATA'
>>> errno.errorcode[errno.ENODATA]
  File "<stdin>", line 1, in <module>
AttributeError: module 'errno' has no attribute 'ENODATA'
```

This looks OS dependent value :-(
```
(venv37zephyr) uname -a
FreeBSD hexagon 12.2-RELEASE-p6 FreeBSD 12.2-RELEASE-p6 GENERIC  amd64
(venv37zephyr) less /usr/include/errno.h
(venv37zephyr) grep ENODATA /usr/include/errno.h
(venv37zephyr) ls -al /usr/include/errno.h
lrwxr-xr-x  1 root  wheel  11 23 lis  2018 /usr/include/errno.h -> sys/errno.h
```

This problem occurs here https://github.com/pyocd/pyOCD/blob/e8797c61083d320a665cd5832ee346d8e074b3ae/pyocd/probe/pydapaccess/interface/pyusb_backend.py#L99 when pyOCD wants to detach kernel driver that may interfere with its operations. Another problem here is that standard user is not supposed to attach/detach kernel drivers. I did put try/except around that and when kernel is not detached things work as expected anyway. 

Now I can see that I have considered fix working because different error occurs and only `except (NotImplementedError, usb.core.USBError) as e:` is handled:
```
Type "help", "copyright", "credits" or "license" for more information.
>>> import errno
>>> errno.ENODATA
  File "<stdin>", line 1, in <module>
AttributeError: module 'errno' has no attribute 'ENODATA'
>>> from errno import ENODATA
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'ENODATA' from 'errno' (unknown location)
```

How can we fix that? How can we fix that on BSD that already has no value defined?

Looks again like no change is necessary on `pyusb` side but the client application..? :-)

cederom: * Okay so after some analysis I can see that this part of `pyusb` may need some improvement.
* Current `pyusb/backend/libusb0.py` code is tied to Linux, which is one problem. `libusb1.py` does not use `ENODATA` at all.
* Another problem is that import is not handled in a way it could raise an exception, so I cannot handle it anyhow from the application level, everything just crashes. 
* FreeBSD implementation differs from Linux implementation and somehow we are passing the `if not hasattr(_lib, 'usb_get_driver_np')` check so it means the function is implemented. Can you please hint us @hselasky on what it returns - is it `usbfs` when in use or `ENODATA` alternative when not used by kernel module?
* Physical interface is running DAPLink firmware - it uses HID (CMSIS-DAP Debug that we are interested in this part), VCP (Serial Console Port to microcontroller), and UMS (drag-n-drop firmware flashing).

jonasmalacofilho: Actually, this is not the only problem: our implementation assumed `usb_get_driver_np()` would only be available on Linux, and makes no sense for FreeBSD's libusb0.

FreeBSD's `usb_get_driver_np()` [appears](https://svnweb.freebsd.org/base/head/lib/libusb/usb.h?revision=326219&view=markup) to simply wrap `libusb20_dev_get_iface_desc()`, which in turn is [documented](https://www.unix.com/man-page/freebsd/3/libusb20_dev_get_iface_desc/) to do:

> `libusb20_dev_get_iface_desc()` retrieves the kernel interface description for the given USB `iface_index`.  The format of the USB interface description is: `"drivername<unit>: <description>"` The description string is always zero terminated.  A zero length string is written in case no driver is attached to the given interface.  The USB device given by `pdev` must be opened before this function will succeed.  This function returns zero on success else a `LIBUSB20_ERROR` value is returned.

Additionally, the `usbfs` driver is a special case on Linux: while it's a kernel driver, it's the one used by libusb.  As that's not a driver that needs to be detached, and following the libusb1 API, our function returns `False` in that case.  A similar consideration may apply to FreeBSD...

cederom: I have tried to enforce libusb1 backend (https://github.com/pyocd/pyOCD/blob/e8797c61083d320a665cd5832ee346d8e074b3ae/pyocd/probe/pydapaccess/interface/pyusb_backend.py#L73) but that does not seem to work and this could save us some work. Note that LibUSB-1.0 is supported on FreeBSD.
```
 72         # Get device handle
 73         if "BSD" in platform.system():
 74             import usb.backend.libusb1 as libusb1
 75             b = libusb1.get_backend()
 76             if b is None:
 77                 print("LIBUSB1 BACKEND UNAVAILABLE!")
 78                 exit(-1)
 79             dev = usb.core.find(custom_match=FindDap(self.serial_number),
 80                 backend=b)
 81         else:
 82             dev = usb.core.find(custom_match=FindDap(self.serial_number))
```
And the result is:
```
(venv37zephyr) pudb3 -m pyocd flash ../zephyr.hex
LIBUSB1 BACKEND UNAVAILABLE!
```
or without printf we can see that libusb0 is selected anyway returning us to the starting point:
```
(venv37zephyr) pudb3 -m pyocd flash ../zephyr.hex
0001962:CRITICAL:__main__:uncaught exception: cannot import name 'ENODATA' from 'errno' (unknown location)
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/__main__.py", line 150, in run
    status = cmd.invoke()
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/subcommands/flash_cmd.py", line 86, in invoke
    with session:
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/core/session.py", line 343, in __enter__
    self.open()
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/core/session.py", line 457, in open
    self._probe.open()
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/probe/cmsis_dap_probe.py", line 170, in open
    self._link.open()
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/utility/concurrency.py", line 28, in _locking
    return func(self, *args, **kwargs)
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/probe/pydapaccess/dap_access_cmsis_dap.py", line 635, in open
    self._interface.open()
  File "/mnt/stuff/x/cederom-pyocd.git/pyocd/probe/pydapaccess/interface/pyusb_backend.py", line 108, in open
    if dev.is_kernel_driver_active(interface_number):
  File "/home/x/usr/local/venv37zephyr/lib/python3.7/site-packages/pyusb-1.1.1-py3.7.egg/usb/core.py", line 1100, in is_kernel_driver_active
    interface)
  File "/home/x/usr/local/venv37zephyr/lib/python3.7/site-packages/pyusb-1.1.1-py3.7.egg/usb/backend/libusb0.py", line 624, in is_kernel_driver_active
    from errno import ENODATA
ImportError: cannot import name 'ENODATA' from 'errno' (unknown location)
```
Help :-)

jonasmalacofilho: What is the `.so` name for FreeBSD's libusb1?

cederom: I took a look at libusb-0.1.12 and there is no `libusb_kernel_driver_active()` yet, probably this is why `usb_get_driver_np` is used? But we have that new function on FreeBSD. Why detection of `libusb1` backend fails?
```
% grep libusb_kernel_driver_active /usr/include/*usb*
/usr/include/libusb.h:int       libusb_kernel_driver_active(libusb_device_handle * devh, int interface);
```

jonasmalacofilho: It fails because PyUSB (using `ctypes.util.find_library()`) can't find the appropriate `.so`.

What is the `.so` name for FreeBSD's libusb1?

cederom: > What is the `.so` name for FreeBSD's libusb1?

```
(venv37zephyr) ls /usr/lib/*usb*
/usr/lib/libusb_p.a     /usr/lib/libusb.so      /usr/lib/libusbhid_p.a  /usr/lib/libusbhid.so
/usr/lib/libusb.a       /usr/lib/libusb.so.3    /usr/lib/libusbhid.a    /usr/lib/libusbhid.so.4
```

cederom: > FreeBSD's `usb_get_driver_np()` [appears](https://svnweb.freebsd.org/base/head/lib/libusb/usb.h?revision=326219&view=markup) to simply wrap `libusb20_dev_get_iface_desc()`, which in turn is [documented](https://www.unix.com/man-page/freebsd/3/libusb20_dev_get_iface_desc/) (..)

Yup :-) There are two variants of this function:
1. `usb_get_driver_np` contained in `libusb01.c` calls `libusb20_dev_get_iface_desc()` and truncates the result (reference: https://cgit.freebsd.org/src/tree/lib/libusb/libusb01.c#n981)
2. `libusb_get_driver_np` contained in `libusb10.c` simply wraps the `libusb_get_driver()` (reference: https://cgit.freebsd.org/src/tree/lib/libusb/libusb10.c#n871).

> Additionally, the `usbfs` driver is a special case on Linux: while it's a kernel driver, it's the one used by libusb. As that's not a driver that needs to be detached, and following the libusb1 API, our function returns `False` in that case. A similar consideration may apply to FreeBSD...

Yes, except this Linux `ENODATA` return code which on FreeBSD seems to be simply `0` or `-1`. Maybe a simple if wrap can be applied here to distinguish the import + response?

If we solve this `libusb0` issue it would be also nice to solve `libusb1` issue and use new library as the default backend.. can we do that without `libusb_get_parent()`?

cederom: Btw. when I `python setup.py install` I get package stripped out of source code and I am unable to debug modified `pyusb` anymore - how can I solve that? :-)

jonasmalacofilho: > Yup :-) There are two variants of this function:

I wouldn't call them variants.  The first is part of the cross-platform libusb0 API, the second is something specific to FreeBSD's libusb1.

> Yes, except this Linux ENODATA return code which on FreeBSD seems to be simply 0 or -1.

From the documentation of `libusb20_dev_get_iface_desc()`, I think the return value can be computed solely from the returned string (taking into account possible usbfs-like behavior).  And, unlike on Linux, any errors can simply be forwarded to the caller (as `USBError`).

Can you prepare and test a patch for this?

> Maybe a simple if wrap can be applied here to distinguish the import + response?

Not sure what you meant there.

> If we solve this libusb0 issue it would be also nice to solve libusb1 issue and use new library as the default backend

Sure, we should fix that too.

> can we do that without `libusb_get_parent()`?

I think so.

By the way, I missed your first mention to `libusb_get_parent()` because I had already answered to a previous version of that comment.  Try to avoid edits that add relevant/brand new information, especially in fairly active issues.

jonasmalacofilho: > Btw. when I `python setup.py install` I get package stripped out of source code and I am unable to debug modified `pyusb` anymore - how can I solve that? :-)

Do you mean with [`pdb`](https://docs.python.org/3/library/pdb.html)?

And have you installed this "modified" PyUSB or are you having problems importing it instead of the installed version?

cederom: > > Yes, except this Linux ENODATA return code which on FreeBSD seems to be simply 0 or -1.
> From the documentation of `libusb20_dev_get_iface_desc()`, I think the return value can be computed solely from the returned string (taking into account possible usbfs-like behavior). And, unlike on Linux, any errors can simply be forwarded to the caller (as `USBError`).

Here is the source:
```
usb_get_driver_np(usb_dev_handle * dev, int interface, char *name, int namelen)
{
	struct libusb20_device *pdev;
	char *ptr;
	int err;

	pdev = (void *)dev;

	if (pdev == NULL)
		return (-1);
	if (namelen < 1)
		return (-1);
	if (namelen > 255)
		namelen = 255;

	err = libusb20_dev_get_iface_desc(pdev, interface, name, namelen);
	if (err != 0)
		return (-1);

	/* we only want the driver name */
	ptr = strstr(name, ":");
	if (ptr != NULL)
		*ptr = 0;

	return (0);
}
```

> Can you prepare and test a patch for this?

Will do :-) Just need to take some sleep :-)

> > If we solve this libusb0 issue it would be also nice to solve libusb1 issue and use new library as the default backend
> Sure, we should fix that too.

:-)


> > can we do that without `libusb_get_parent()`?
> I think so.

:-)

> By the way, I missed your first mention to `libusb_get_parent()` because I had already answered to a previous version of that comment. Try to avoid edits that add relevant/brand new information, especially in fairly active issues.

ACK :-)

cederom: > > Btw. when I `python setup.py install` I get package stripped out of source code and I am unable to debug modified `pyusb` anymore - how can I solve that? :-)
> 
> Do you mean with [`pdb`](https://docs.python.org/3/library/pdb.html)?

Well I am using ncurses based `pudb3` (https://pypi.org/project/pudb/) but this is the same + text gui for faster navigation (like TUI+GDB / CGDB). Instead `import pdb; pdb.set_trace()` I am calling `import pudb; pudb.set_trace()`.

> And have you installed this "modified" PyUSB or are you having problems importing it instead of the installed version?

I am working in Python VirtualEnv, usually I just install packages with `pip install pyusb` but for testing I am removing them with `pip uninstall pyusb` and then I install modified version for testing with `python setup.py install`. I need to install `pyusb` as this is required package for `pyocd`. If package is no dependency for others I can simply call it with `pudb3 -m pyocd` directly from the source code repository :-)

cederom: `python setup.py develop` is the answer in place of `python setup.py install` - that will install development version of module with debug symbols included :-)

cederom: > > Can you prepare and test a patch for this?

Done :-) Please verify @jonasmalacofilho :-)

usb/backend/libusb0.py
```diff
@@ -619,27 +619,47 @@ def reset_device(self, dev_handle):
 
     @methodtrace(_logger)
     def is_kernel_driver_active(self, dev_handle, intf):
+        # libusb0's usb_get_driver_np() is platform specific / not portable.
+        # This function checks what kernel module handles dev_handle/intf.
+        # Various implementations may return system specific values - for
+        # instance Linux may return errno.ENODATA not present on FreeBSD.
+        # On FreeBSD successul call returns 0 otherwise -1.
+
```
jonasmalacofilho, line 627: This goes in a bit too much detail, but also isn't clear enough.  Remove all but the first lines of this comment...

usb/backend/libusb0.py
```diff
@@ -619,27 +619,47 @@ def reset_device(self, dev_handle):
...
         buf = usb.util.create_buffer(_USBFS_MAXDRIVERNAME + 1)
         name, length = buf.buffer_info()
         length *= buf.itemsize
-        # based on the implementation of libusb_kernel_driver_active
+
+        # Linux: based on the implementation of libusb_kernel_driver_active
         # (see libusb/os/linux_usbfs.c @@ op_kernel_driver_active):
         # usb_get_driver_np fails with ENODATA when no kernel driver is bound,
         # and if 'usbfs' is bound that means that a userspace program is
         # controlling the device (e.g. using this very library)
```
jonasmalacofilho, line 638: Instead, add a corresponding `# FreeBSD: ...` section here.

usb/backend/libusb0.py
```diff
@@ -619,27 +619,47 @@ def reset_device(self, dev_handle):
...
+        if "linux" in sys.platform:
+            try:
+                from errno import ENODATA
+            except:
+                raise NotImplementedError(self.is_kernel_driver_active.__name__)
+
```
jonasmalacofilho, line 644: This is not necessary, `errno` is already imported at the top of the module, and you're already doing `errno.ENODATA` bellow.

usb/backend/libusb0.py
```diff
@@ -619,27 +619,47 @@ def reset_device(self, dev_handle):
...
                         intf,
                         cast(name, c_char_p),
                         length))
-            return cast(name, c_char_p).value != b'usbfs'
+            if "linux" in sys.platform:
+                # On Linux 'usbfs' indicates userland application.
+                return cast(name, c_char_p).value != b'usbfs'
+            else:
+                # On other platforms just check any module attached.
+                return buf.index(0) != 0
```
jonasmalacofilho, line 656: Have you confirmed that this behaves as `libusb_kernel_driver_active()`?

> Determine if a kernel driver is active on an interface. If a kernel driver is active, you cannot claim the interface, and libusb will be unable to perform I/O.

Specifically, `is_kernel_driver_active()` should return `False` if the kernel driver bound to the device is the one this backend uses (or will use anyway).

This is counter intuitive given the function name, but the idea is that this method, like the equivalent libusb-1 API, is used to check whether something has to be done about the current kernel driver before being able to interact with the device.

jonasmalacofilho (changes requested):

usb/backend/libusb0.py
jonasmalacofilho: Thanks!

jonasmalacofilho: Also, I really need to stop using GitHub's "request changes" option when submitting a review.  While its description is more or less accurate,

> feedback that must be addressed before merging.

I think that's only visible to me, and I'm obviously not _requesting_ anything.  Instead, I'm _asking_ for you to make these changes.

: )

cederom: no worries, whatever button, we press it :-P

usb/backend/libusb0.py
cederom:
```
622         # libusb0's usb_get_driver_np() is platform specific / not portable.
623         # This function checks what kernel module handles dev_handle/intf.
624         # Various implementations may return system specific values.
```
?

usb/backend/libusb0.py
cederom:
```
637         # FreeBSD: ENODATA is undefied, returns 0 on success, -1 on error.
```
?

usb/backend/libusb0.py
cederom: Yes, this is additional check, `ENODATA` may not be defined on (some older) Linux (?), this will be explicitly inidcated with  `NotImplementedError` exception, instead throwing unhandled exception later :-)

usb/backend/libusb0.py
cederom: * Cannot say for every OS but on FreeBSD we can just write do /dev/usb*` from the user standpoint if file permission allows it. 
* User should not be allowed to load/unload kernel modules, only root can do that, so unloading kernel module will always fail when performed by user application. 
* This is why in `pyOCD` failed unloading module will drop a warning instead application failure. Then it tried to write to a device anyway and it works just fine.
* On FreeBSD `uhid` kernel module is attached to the `CMSIS-DAP` interface.

From what I can see only FreeBSD is using its own LibUSB implementation that "joins" 0.1 and 1.0 functionalities. OpenBSD has `libusb-1.0` package and has no `ENODATA` defined. NetBSD has `ENODATA` defined and `libusb-0.1.12nb7`, `libusb-compat-0.1.7`, `libusb1-1.0.24`, as well as `py38-libusb1-1.7.1` and `libusb-0.1.12nb7` packages. But I guess their behavior is similar - user can just write to `/dev/usb*` [1][2][3]. I only have VM with OpenBSD and NetBSD if necessary I can verify these functions on real hardware..?

Is this correct @hselasky ? :-)

I have tried to get a common denominator here :-)

[1] https://nixdoc.net/man-pages/FreeBSD/man4/usb.4.html
[2] https://man.openbsd.org/usb.4
[3] https://man.netbsd.org/usb.4

cederom:
Updated proposition pushed for review :-)

cederom:
I just want to let you know that I remember about this task. I started writing libusb example to show how this specific function behaves on all OS on all devices and interfaces. That will be best test and simplest way of verification. Maybe something similar written in `pyusb`.
However, I have several fires to extinguish at the moment, and new shit hits the fan again and again, so I have to focus on priorities :-( Will get back here in a free moment. If anyone has free moment to test changes please go on and verify :-)

jonasmalacofilho:
Thanks for the letting me know, and that's absolutely fine.

When you have some time, can you just find out the answers for these questions:

> What does `usb_get_driver_np()` return in `buf` (on FreeBSD) in all possible cases:
> 
>     * No kernel driver bound to the device?
>     * Kernel driver active, but it's the one used by this backend (=libusb0)?
>     * Other kernel driver active (=this backend will not be able to access the device)?

That should be enough for someone like me to pick up the task.

mcuee:
Yeah FreeBSD libusb is a bit different from other OS, it is not done by the upstream libusb project but done by FreeBSD developers.

HPS (Hans Petter Selasky) was the main author of the "new" FreeBSD USB Stack and libusb implementation). I still remembered I was tetsing the USB Stack implementation by building FreeBSD kernel and "World" many years ago. I was also testing the FreeBSD libusb implementation at that time. Then I was testing pyusb on FreeBSD as well. Time flies.

mcuee:
libusb wiki:
https://github.com/libusb/libusb/wiki#Supported_Environments

Note: FreeBSD 8 and above include a FreeBSD-specific reimplementation of the libusb API, so your applications will probably work there too. The source code for this library can be found here (https://svnweb.freebsd.org/base/head/lib/libusb/).

mcuee:
@hselasky 

Please help to clarify the following. Thanks.
> What does `usb_get_driver_np()` return in `buf` (on FreeBSD) in all possible cases:
> 
>     * No kernel driver bound to the device?
>     * Kernel driver active, but it's the one used by this backend (=libusb0)?
>     * Other kernel driver active (=this backend will not be able to access the devic

hselasky:
> @hselasky
> Please help to clarify the following. Thanks.
> 
> I remember FreeBSD allows libusb to access certain generic USB device by binding both ugen driver and specific kernel driver to a USB device.
> 
> > What does `usb_get_driver_np()` return in `buf` (on FreeBSD) in all possible cases:
> > ```
> > * No kernel driver bound to the device?

It returns a negative failure error code, -1. (ENODATA is only for Linux I believe)

> > * Kernel driver active, but it's the one used by this backend (=libusb0)?

You must detach active kernel drivers. There is no mechanism in FreeBSD to prevent libusb and kernel drivers from accessing the same endpoints. See below.

> > * Other kernel driver active (=this backend will not be able to access the device.

It returns for example uaudioX in buf, and 0 for return value.

> > ```

Hope this helps.

--HPS

jonasmalacofilho:
@hselasky,

Thanks!

> It returns a negative failure error code, -1. (ENODATA is only for Linux I believe)

Hm, can you confirm that's really the case?

I [only] ask because FreeBSD's `usb_get_driver_np()` [appears](https://svnweb.freebsd.org/base/head/lib/libusb/usb.h?revision=326219&view=markup) to simply wrap `libusb20_dev_get_iface_desc()`, and in turn its [documentation](https://www.unix.com/man-page/freebsd/3/libusb20_dev_get_iface_desc/) states that:

> `libusb20_dev_get_iface_desc()` retrieves the kernel interface description [...]  **A zero length string is written in case no driver is attached to the given interface.** [...]

---

> You must detach active kernel drivers. There is no mechanism in FreeBSD to prevent libusb and kernel drivers from accessing the same endpoints. See below.

But doesn't libusb itself use a kernel driver to access the device?

And, no matter the mechanism, what does `usb_get_driver_np()` return while libusb has successfully gained control of the interface?

(My concern is that, even if libusb uses some other mechanism than a driver to access devices, `usb_get_driver_np()` could still return some magic string to indicate when this is going on).

Thanks again.

mcuee:
> But doesn't libusb itself use a kernel driver to access the device?

I think FreeBSD libusb uses ugen to access USB device (OpenBSD and NetBSD also use ugen). Similarly libusb under Linux use usbfs to access a USB device. 

Technically ugen and usbfs are kernel driver as well, but in the case of libusb, they are considered as "not using other kernel driver". In the case of Linux, if a kernel driver is detached, then usbfs is the driver used and then libusb is able to access it. Probably it is the same that ugen will be used when the other kernel driver is detached.

hselasky:
@jonasmalacofilho

You are right. I was a bit too quick there.

usb_get_driver_np() returns a non-zero or negative value when the interface index is invalid or some of the parameters are toast.

Else:

Non-zero length string if kernel driver is attached

Else

Zero-length string and 0 return code if no kernel driver is attached.

Does that make sense. The manual page is correct apparently.

--HPS

mcuee:
Yes looks like the documentation is correct.

>
libusb20_dev_get_iface_desc() retrieves the kernel interface description for the given USB iface_index.  The format of the USB interface
     description is: "drivername<unit>: <description>" The description string is always zero terminated.  A zero length string is written in case
     no driver is attached to the given interface.  The USB device given by pdev must be opened before this function will succeed.  This function
     returns zero on success else a LIBUSB20_ERROR value is returned.

hselasky:
@mcuee: In FreeBSD user-space drivers like webcamd, create a uniq PID file, and exit if there is already a PID file existing for the USB device.

Else FreeBSD is probably a bit more flexible than Linux. There is currently no way to query the kernel if another libusb application is using the device, except may query the system for open files, and you'll see.

hselasky:
@mcuee 
> I think FreeBSD libusb uses ugen to access USB device (OpenBSD and NetBSD also use ugen). Similarly libusb under Linux use usbfs to access a USB device.

Ugen in FreeBSD has some IOCTL extensions which makes it work more like Linux's USBFS. NetBSD and OpenBSD only allow one endpoint per ugen node, if I'm not mistaken. While FreeBSD allows multiple endpoint nodes per UGEN file handle.

--HPS

mcuee:
> @mcuee: In FreeBSD user-space drivers like webcamd, create a uniq PID file, and exit if there is already a PID file existing for the USB device.
> 
> Else FreeBSD is probably a bit more flexible than Linux. There is currently no way to query the kernel if another libusb application is using the device, except may query the system for open files, and you'll see.

I see. Thanks. Typical libusb application under Linux will carry out the following. It will try to detach the kernel driver first and then claim the interface. If interface claiming failed (another libusb application is claiming the interface), then the libusb application will fail and exit.

cederom:
Thank you everyone for feedback and information! :-)

> I see. Thanks. Typical libusb application under Linux will carry out the following. It will try to detach the kernel driver first and then claim the interface. If interface claiming failed (another libusb application is claiming the interface), then the libusb application will fail and exit.

On FreeBSD at the moment I am using DAPLink debug probe + pyOCD as probe controller and GDB remote backend. DAPLink uses CMSIS-DAP interface for accessing debug chip stuff which uses HID for USB transport. HID is usually handled by `uhid` module on FreeBSD. However, even if `uhid` is attached to CMSIS-DAP interface, I get this `uhid2` string returned by `usb_get_driver_np()`, I do not fail pyOCD on failed detach attempt (because as standard user I am not allowed to manipulate kernel modules), and things just work fine - with attached `uhid` module and `pyusb` operating on that interface.

mcuee:
Just a side note here for libusb-1.0 under macOS:
libusb_kernel_driver_active() is available under macOS but may be broken now. I am not so sure why it is available under macOS.
https://github.com/libusb/libusb/issues/908

mcuee:
> Thank you everyone for feedback and information! :-)
> 
> > I see. Thanks. Typical libusb application under Linux will carry out the following. It will try to detach the kernel driver first and then claim the interface. If interface claiming failed (another libusb application is claiming the interface), then the libusb application will fail and exit.
> 
> On FreeBSD at the moment I am using DAPLink debug probe + pyOCD as probe controller and GDB remote backend. DAPLink uses CMSIS-DAP interface for accessing debug chip stuff which uses HID for USB transport. HID is usually handled by `uhid` module on FreeBSD. However, even if `uhid` is attached to CMSIS-DAP interface, I get this `uhid2` string returned by `usb_get_driver_np()`, I do not fail pyOCD on failed detach attempt (because as standard user I am not allowed to manipulate kernel modules), and things just work fine - with attached `uhid` module and `pyusb` operating on that interface.

By right you should hidapi and its python binding for HID compatible DAP-Link. But again hidapi under FreeBSD will need to use libusb so it does not really offer advantage over libusb/pyusb.

mcuee:
@cederom Another thing is that it is a bit strange that pyOCD does not work with libusb-1.0 backend. That should be the preferred backend to use compared to libusb-0.1.

mcuee:
From what I read libusb0.py is a bit strange compared to libusb1.py in terms of dealing with the kernel driver thingy. 

Maybe libusb0.py is trying to do too much, probably due to the limitation of libusb-0.1 API. 

libusb1.py seems to be simple -- just wrap the libusb API.

usb/backend/libusb0.py
mcuee: OpenBSD also has libusb-1.0 and libusb-compat.
http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/devel/libusb1/
http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/devel/libusb-compat/

The libusb-0.1 is obsolete in OpenBSD.
http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/devel/libusb/

mcuee: NetBSD have libusb-1.0.
https://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/devel/libusb1/

NetBSD keeps both libusb-compat-0.1 and legacy libusb-0.1
https://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/devel/libusb-compat/
https://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/devel/libusb/

jonasmalacofilho:
Based on the comments on this PR, I tried to implement a new (and hopefully complete) fix for `is_kernel_driver_active()` on FreeBSD.

Can you please review and test the code in my [`fix-libusb0-is-kernel-driver-active`](https://github.com/jonasmalacofilho/pyusb/tree/fix-libusb0-is-kernel-driver-active) tree ([diff](https://github.com/pyusb/pyusb/compare/master...jonasmalacofilho:fix-libusb0-is-kernel-driver-active?expand=1))?

hselasky:
Looks good from the FreeBSD's point of view.

mcuee:
The changes (fix-libusb0-is-kernel-driver-active ) look good.

jonasmalacofilho:
@cederom have you had a chance to take a look at or try the alternative fix?  Do you have any objections (empirical or theoretical)?

cederom:
hello world. skydiving this weekend. will check on monday. thank you! =)

--
CeDeROM, SQ7MHZ, http://www.tomek.cedro.info

cederom:
Im back :-) Weekend was fun, week like jumping out with no parachute, sorry for delay, Im on it :-)

image

I will now verify fix by @jonasmalacofilho and the `ENODATA` import.

jonasmalacofilho:
@cederom,

Thanks!

> Im back :-) Weekend was fun, week like jumping out with no parachute, sorry for delay, Im on it :-)

Yeah, I'm also having one of those weeks.

PR #412, opened 2022-03-02
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help automation tool find the source code for Requests.

Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,9 @@ setup(
     author='Jonas Malaco',
     author_email='me@jonasmalaco.com',
     url='https://pyusb.github.io/pyusb',
+    project_urls={
+        'Source': 'https://github.com/pyusb/pyusb',
+    },
     packages=['usb', 'usb.backend'],
     long_description=
 """


jonasmalacofilho:
Looks good to me, and it seems probable that "Source" will indeed be the recommended name, or at least one of them (more info in pypa/warehouse#5947).

(Note: right now, the icon would be selected not because of "Source", but because the URL [would contain "github.com"](https://github.com/pypa/warehouse/blob/8921520cfee911bd6c8b29ad1160280b2043416d/warehouse/templates/packaging/detail.html#L32-L33)).

---

Since you mentioned it, can you should me how this looks in an API response?

Thanks!

aoriekhov-godaddy:
> Since you mentioned it, can you should me how this looks in an API response?

https://pypi.org/pypi/requests/json

jonasmalacofilho:
Alright! Thanks again!

PR #413, opened 2022-03-07
Loads libusb dll from the working directory

JackLilhammers:
Hello,
while working on a cross plaftorm app I had some issues trying to run PyUSB on Windows, because I tried to avoid installing libusb binaries in system32.
Linux comes with libusb preinstalled and on macOS if I put the dylib installed with Homebrew in the working directory it PyUSB runs fine.
These changes allow PyUSB to use a libusb dll placed in the working directory on Windows , just like in macOS

### DIFF
--- a/usb/libloader.py
+++ b/usb/libloader.py
@@ -94,13 +94,18 @@ def locate_library (candidates, find_library=ctypes.util.find_library):
         # Workaround for CPython 3.3 issue#16283 / pyusb #14
         if use_dll_workaround:
             candidate += '.dll'
-            libpath = lambda p: os.path.abspath(p)
-        else:
-            libpath = lambda p: p
 
         libname = find_library(candidate)
         if libname:
-            return libpath(libname)
+            return libname
+
+        # If the library isn't found in the system path
+        # looks for it in the working directory
+        if use_dll_workaround:
+            candidate = os.path.abspath(candidate)
+            libname = find_library(candidate)
+            if libname:
+                return libname
     # -- end for
     return None
 
--- a/usb/libloader.py
+++ b/usb/libloader.py
@@ -67,7 +67,7 @@ class LibraryMissingSymbolsException(LibraryException):
     pass
 
 
-def locate_library (candidates, find_library=None):
+def locate_library (candidates, find_library=ctypes.util.find_library):
     """Tries to locate a library listed in candidates using the given
     find_library() function (or ctypes.util.find_library).
     Returns the first library found, which can be the library's name
--- a/usb/libloader.py
+++ b/usb/libloader.py
@@ -33,6 +33,7 @@
 import ctypes
 import ctypes.util
 import logging
+import os
 import sys
 
 __all__ = [
@@ -66,7 +67,7 @@ class LibraryMissingSymbolsException(LibraryException):
     pass
 
 
-def locate_library (candidates, find_library=ctypes.util.find_library):
+def locate_library (candidates, find_library=None):
     """Tries to locate a library listed in candidates using the given
     find_library() function (or ctypes.util.find_library).
     Returns the first library found, which can be the library's name
@@ -93,10 +94,13 @@ def locate_library (candidates, find_library=ctypes.util.find_library):
         # Workaround for CPython 3.3 issue#16283 / pyusb #14
         if use_dll_workaround:
             candidate += '.dll'
+            libpath = lambda p: os.path.abspath(p)
+        else:
+            libpath = lambda p: p
 
         libname = find_library(candidate)
         if libname:
-            return libname
+            return libpath(libname)
     # -- end for
     return None
 


usb/libloader.py
```diff
@@ -93,10 +94,13 @@ def locate_library (candidates, find_library=ctypes.util.find_library):
...
         if use_dll_workaround:
             candidate += '.dll'
+            libpath = lambda p: os.path.abspath(p)
+        else:
+            libpath = lambda p: p
 
         libname = find_library(candidate)
         if libname:
-            return libname
+            return libpath(libname)
```
jonasmalacofilho, line 103: So this means that `ctypes.util.find_library()` already finds and returns the path to a local DLL, but it's a relative path... is that right?

Additionally, is it specifically the path to a DLL in the working directory, the Python interpreter directory, or some other local (i.e. non-system) directory?

usb/libloader.py
```diff
@@ -66,7 +67,7 @@ class LibraryMissingSymbolsException(LibraryException):
     pass
 
 
-def locate_library (candidates, find_library=ctypes.util.find_library):
```
jonasmalacofilho, line 69: This looks redundant with

```py
    if find_library is None:
        find_library = ctypes.util.find_library
```

but isn't _in intention_.  At least I think it isn't (I'm not the original author).

Technically, the conditional alone would be enough, of course, but having an obvious default is valuable...  Probably more than supporting the case where the caller is unable to decide whether to include the parameter or not, thus relying on `None` to signal "the default, please".

I'm leaning towards keeping the minor redundancy in favor or expressiveness.  Do you mind reverting this part?

JackLilhammers: No problem

usb/libloader.py
JackLilhammers: I worked on this on my work pc and in the end a library must have remained somewhere and everything ran fine even if my PR was buggy.
Now I've tested and fixed it on my laptop and it seems better.

When ctypes.util.find_library() doesn't find the library in the system path, I try again looking for it in the program working directory

jonasmalacofilho: > When ctypes.util.find_library() doesn't find the library in the system path, I try again looking for it in the program working directory

Oh, I see.  This could deviate from what's expected, security-wise, on those platforms.  I'm going to need to look into that in more detail... I'll try to do it in the next few days.

JackLilhammers: You mean on Windows?

Anyway, thank you!

jonasmalacofilho: You're right, `use_dll_workaround` makes sure this is only changing behavior there.

And if anyone asked me yesterday what the default behavior on Windows is, I'd have said to load from the current working directory without issues.  But the fact that `ctypes.util.find_library()` doesn't already do this is surprising, as IIRC it just uses whatever functionality the platform already supports.  So I'm probably wrong about one of those assumptions...

JackLilhammers: I'd have said the same thing, but now I've found that Python 3.8 changed this behaviour for Windows
[Changes in the Python API](https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-python-api)

> DLL dependencies for extension modules and DLLs loaded with [ctypes](https://docs.python.org/3.8/library/ctypes.html#module-ctypes) on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with [add_dll_directory()](https://docs.python.org/3.8/library/os.html#os.add_dll_directory) are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for [add_dll_directory()](https://docs.python.org/3.8/library/os.html#os.add_dll_directory) and if it exists, use it to add your DLLs directory while loading your library. Note that Windows 7 users will need to ensure that Windows Update KB2533623 has been installed (this is also verified by the installer). (Contributed by Steve Dower in [bpo-36085](https://bugs.python.org/issue36085).)

I tried
```python
if use_dll_workaround:
    candidate += '.dll'
    if sys.version_info.major >= 3 and sys.version_info.minor >= 8:
        with os.add_dll_directory(os.getcwd()) as cwd:
            libname = find_library(candidate)
else:
    libname = find_library(candidate)
```
but it doesn't seem to work...

This is the mre
```python
import usb.core

dev = usb.core.find()

print(dev)
```

jonasmalacofilho: Thanks for looking into it, that's exactly what I was worried about.

There are already ways to solve the loading libusb problem when you're (directly) distributing software to WIndows users, a scenario where you can make sure that the extra location is still secure):

- pass a custom `find_library` function or lambda;
- use `add_dll_directory()`.

_(I think the latter should only be used in the application that's directly distributed and installed by Windows users; using it anywhere upstream from that would silently open security holes.  The former I could see being used in an intermediate library, something sitting between PyUSB and the application, but it should be as specific as possible, possibly only loading DLLs that it [the intermediate library] ships itself)._

But the bottom line is: we're not going to silently and broadly override/undermine a security policy in Python.

So I'm sorry, but I can't take this change.  Still, thanks for proposing/discussing it.

JackLilhammers: Makes sense, but then maybe it could be added to the readme or something like that, because we both thought it worked differently

Something like:
To load libusb dll on Windows you can run

```python
import os
import sys
import usb.backend.libusb1

if sys.platform == 'win32':
    with os.add_dll_directory(os.getcwd()):
        usb.backend.libusb1.get_backend()
```

at the beginning of your module

jonasmalacofilho: It would need to be a lot more carefully worded than that; specially, that should only be very last resort, and definitively not something libraries that depend on PyUSB should ever do.

But there's a better way: [pyocd/libusb-package](https://github.com/pyocd/libusb-package).  I'm experimenting with it now, and later will audit how the wheels are built.  Then I plan on adding it to our docs, and _maybe_ also have `usb.core.find()` use it automatically, when available, as a fallback (although this last still requires some thought).

PR #479, opened 2023-05-18
Fix bMaxPower conversion for USB 3 devices

andygoulden:
As per the 'FIXME' comment, the current value uses a different multiplier for USB 3 (ie superspeed) devices.

This change uses the USB version from bcdUSB to determine the correct multiplier.

### DIFF
--- a/usb/core.py
+++ b/usb/core.py
@@ -671,7 +671,7 @@ class Configuration(object):
 
     def _get_power_multiplier(self):
         if self.device.speed is not None:
-            power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if self.device.speed == 3 else _lu.MAX_POWER_UNITS_USB2p0
+            power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if self.device.speed >= 4 else _lu.MAX_POWER_UNITS_USB2p0
         else:
             power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if self.device.bcdUSB >= 0x0300 else _lu.MAX_POWER_UNITS_USB2p0
 
--- a/usb/core.py
+++ b/usb/core.py
@@ -669,13 +669,18 @@ class Configuration(object):
         """
         return Interface(self.device, index[0], index[1], self.index)
 
+    def _get_power_multiplier(self):
+        if self.device.speed is not None:
+            power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if self.device.speed == 3 else _lu.MAX_POWER_UNITS_USB2p0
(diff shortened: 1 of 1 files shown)


usb/core.py
```diff
@@ -671,9 +671,11 @@ def __getitem__(self, index):
 
 
     def _str(self):
+        power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if ((self.device.bcdUSB & 0xff00)>>8) == 3 else _lu.MAX_POWER_UNITS_USB2p0
```
jonasmalacofilho, line 674: I think we should first check `self.device.speed` (from [`libusb_get_device_speed`](https://libusb.sourceforge.io/api-1.0/group__libusb__dev.html#ga58c4e448ecd5cd4782f2b896ec40b22b)), and only fallback to `bcdUSB` when that isn't available.

Also, I suggest replacing the mask and exact comparison with `self.device.bcdUSB >= 0x0300`.

Finally, the choice of multiplier is duplicated between this function and `_get_full_descriptor_str`, and non trivial; can you make sure it's only done in a single place?

andygoulden: Thanks for your feedback. I've pushed up a commit which addresses your comments - let me know what you think.

usb/core.py
```diff
@@ -669,11 +669,18 @@ def __getitem__(self, index):
         """
         return Interface(self.device, index[0], index[1], self.index)
 
+    def _get_power_multiplier(self):
+        if self.device.speed is not None:
+            power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if self.device.speed == 3 else _lu.MAX_POWER_UNITS_USB2p0
```
jonasmalacofilho, line 674:
```suggestion
            power_multiplier = _lu.MAX_POWER_UNITS_USB_SUPERSPEED if self.device.speed >= 4 else _lu.MAX_POWER_UNITS_USB2p0
```

---

`libusb_supported_speed` constants also form a bitmap, and `LIBUSB_SUPER_SPEED_OPERATION` has value 4, not 3.[^1]

And while some future SuperDuperSpeed mode will likely change the bMaxPower multiplier again (to some value we can't predict), it might be a bit better to (mis)handle that future case as SuperSpeed, instead of <= HighSpeed.

Thanks again!

[^1]: https://github.com/libusb/libusb/blob/cc498ded18fb2c6e4506c546d0351c4ae91ef2cc/libusb/libusb.h#L498

andygoulden: Thanks for the suggestion - I've committed it. Is there anything else that you'd like me to change before it's pulled?

jonasmalacofilho:
Everything looks good to me now, merging...

Thanks again!

PR #495, opened 2023-10-20
legacy: add Bluetooth related constants

antoniovazquezblanco:
Getting in touch with the library. I work with Bluetooth devices, those constants are useful to me.
Those constants may be also useful to others :D


jonasmalacofilho:
LGTM, thanks.

Of course, it's a bit odd to add new stuff to a module called `legacy`. But I also feel that the new constants belong there more than in `util`, so I'm fine with it...

Can you leave a link here (in the PR comments) to the normative reference for the new constants?

Oh, and just for curiosity, are you using the legacy API?

antoniovazquezblanco:
> Oh, and just out of curiosity, are you using the legacy API?

I am using the new API :)

> Of course, it's a bit odd to add new stuff to a module called legacy. But I also feel that the new constants belong there more than in util, so I'm fine with it...

Yep, that was odd haha. In fact I needed to perform a search to find the location of the constants...

I was wondering if there was a better place to put them but I am not familiar with the project :/

> Can you leave a link here (in the PR comments) to the normative reference for the new constants?

https://www.usb.org/defined-class-codes

Thank you!

jonasmalacofilho:
Thanks again!

PR #502, opened 2024-02-29
legacy: Fix `DeviceHandle` finalizer checking for the wrong attribute

puddly:
Introduced in https://github.com/pyusb/pyusb/commit/e658484a03af0e8b3733105c67fe23d76232fc03. I believe this is a typo.

`hasattr(self, 'self.dev')` will never evaluate to `True` so `_finalize_object` will never actually run.


jonasmalacofilho (approved):
Thanks!

Issue #509, opened 2024-04-15
No Backend available MAC


gabri3l0:
Help me i can not usse usb core on mac 14.4
<img width="1269" alt="image" src="https://github.com/pyusb/pyusb/assets/42877617/45e4ea00-c4e9-4523-8f3d-6fb4080bd0e4">

gabri3l0:
Ohh i found it

Summary
run 
`ln -s /opt/homebrew/lib ~/lib`

why
Create a symbolic link in your home directory. This takes advantage of the fact that ~/lib is a default fallback path for libraries:

Thanks @codo @stackoverflow
https://stackoverflow.com/a/70730920/24210494

Issue #361, opened 2021-04-14
pyenv with M1 chip mac cannot find libusb?


ketaro-m:
I'm using M1 mac and Big Sur 11.2.3, python version is 3.8.6 by pyenv.
I installed libusb at first, then created venv and pip-installed pyusb. But when I try using this module, the error like below occurred.

```
In [1]: import usb.core

In [2]: dev = usb.core.find()
---------------------------------------------------------------------------
NoBackendError                            Traceback (most recent call last)
<ipython-input-2-bbeca608b684> in <module>
----> 1 dev = usb.core.find()

~/pyusb/.venv/lib/python3.8/site-packages/usb/core.py in find(find_all, backend, custom_match, **args)
   1297                 break
   1298         else:
-> 1299             raise NoBackendError('No backend available')
   1300
   1301     if find_all:

NoBackendError: No backend available

In [3]:
```

However, if I changed python from pyenv python to homebrew-installed python, the problem above was solved. I guess it's due to the path dependency; homebrew python located in `opt/homebrew/opt/python@3.8` can find `/opt/homebrew/opt/libusb`, but pyenv python located in `~/.pyenv/shims/python` cannot. Is it right?

```
In [1]: import usb.core

In [2]: dev = usb.core.find()

In [3]:
```

This problem doesn't occur with Intel chip mac. Homebrew path is `/usr/local` in Intel processor, `/opt/homebrew` in ARM. 
Could you update this?

jonasmalacofilho:
Basically, we use `ctypes.util.find_library` and that hasn't yet been patched for the 3.8 Python branch (see [Python issue 41100](https://bugs.python.org/issue41100)).  Homebrew's Python 3.8 apparently backports the fix in 3.9, which is why it works.

A bit more information can be found on a previous issue: #355.

Issue #355, opened 2021-01-21
Mac M1 support


2231puppy:
The Mac M1 series seems to be unsupported. Here is my debug output:

```
2021-01-21 16:49:22,365 ERROR:usb.libloader:Libusb 1 (/opt/homebrew/Cellar/libusb/1.0.24/lib/libusb-1.0.dylib) could not be loaded
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/usb/libloader.py", line 118, in load_library
    return ctypes.CDLL(lib)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/opt/homebrew/Cellar/libusb/1.0.24/lib/libusb-1.0.dylib, 6): no suitable image found.  Did find:
	/opt/homebrew/Cellar/libusb/1.0.24/lib/libusb-1.0.dylib: mach-o, but wrong architecture
	/opt/homebrew/Cellar/libusb/1.0.24/lib/libusb-1.0.0.dylib: mach-o, but wrong architecture
	/opt/homebrew/Cellar/libusb/1.0.24/lib/libusb-1.0.dylib: mach-o, but wrong architecture
	/opt/homebrew/Cellar/libusb/1.0.24/lib/libusb-1.0.0.dylib: mach-o, but wrong architecture
2021-01-21 16:49:22,367 ERROR:usb.backend.libusb1:Error loading libusb 1.0 backend
2021-01-21 16:49:22,367 ERROR:usb.libloader:'OpenUSB library' could not be found
2021-01-21 16:49:22,367 ERROR:usb.backend.openusb:Error loading OpenUSB backend
2021-01-21 16:49:22,369 ERROR:usb.libloader:'Libusb 0' could not be found
2021-01-21 16:49:22,369 ERROR:usb.backend.libusb0:Error loading libusb 0.1 backend
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/usb/core.py", line 1299, in find
    raise NoBackendError('No backend available')
usb.core.NoBackendError: No backend available
```

Please add this support.

jonasmalacofilho:
The correct architecture of LibUSB must be installed:

```
[...]

PR #511, opened 2024-04-30
Search in default homebrew prefix on Apple Silicon

jonasmalacofilho:
On Apple Silicon, homebrew will by default (and, really, it's the only recommended way to use it) use the `/opt/homebrew` prefix. The old `/usr/local` prefix is retained for the use of Intel/x86_64 homebrew.

This means that libraries (e.g. LibUSB) installed through homebrew on Apple Silicon will end up in `/opt/homebrew/lib/`. But that location isn't one of the search paths checked by `ctypes.util.find_library` (which is what we use if no custom `find_library` is supplied by the caller). Homebrew [patches](https://github.com/Homebrew/homebrew-core/blob/e91fb0c6e50a/Formula/p/python%403.11.rb#L196-L203) the Python interpreters they distribute, but other/stock interpreters don't know about this additional location for libraries.

As this is a frequent enough issue, let's check for that additional (and hardcoded) prefix if nothing is found in the normal search paths.

Note that, just like `ctypes.util.find_library`, we currently don't take into account the architecture: the first library found with a suitable name is used, even if it doesn't match the architecture of the Python interpreter in use.

This PR also includes basic maintenance (including bumping the minimum supported Python version), mostly just so it can run in our CI.

Related: #355
Related: #361
Related: #509
Related: https://github.com/liquidctl/liquidctl/pull/637#issuecomment-2081156293

### DIFF
--- a/.github/workflows/run_tox.yml
+++ b/.github/workflows/run_tox.yml
@@ -22,11 +22,15 @@ jobs:
         python -m pip install --upgrade pip
         pip install setuptools>=30.3.0 tox
     - name: Install libusb-1.0 (Windows)
-      if: matrix.platform == 'windows-latest'
+      if: startsWith(matrix.platform, 'windows-')
       shell: pwsh
       run: |
         Invoke-WebRequest -Uri https://github.com/libusb/libusb/releases/download/v1.0.24/libusb-1.0.24.7z -OutFile libusb-1.0.24.7z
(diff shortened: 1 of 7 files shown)


clach04:
My 2 cents, mac changes look fine from a quick skim (I'm not a mac user, excellent explanation on why this is needed 👍 ).

PR also changes minimum Python requirements but that's not mentioned in the PR - seems out of scope for fixing mac support.

PR #514, opened 2024-05-13
Type hints for core.py, util.py, control.py

o-murphy:
Added stubs with type hints for core, util., control modules and it's undeclared descriptor's attributes


jonasmalacofilho:
I don't think the level of duplication with this approach is maintainable. And it makes it even harder to check that the type hints are actually correct.

I also spotted a few incorrect or incomplete type hints: e.g. endpoints can sometimes take both `Endpoint` or integer endpoint addresses; array-like data params can be `array`s, lists, iterables, strings, but _not_ `bytearray`s (at least for now).

PR #519, opened 2024-07-09
Search in default homebrew prefix on Apple Silicon fix

mikhail-ekzi:
Continuing the work of the author from the topic https://github.com/pyusb/pyusb/pull/511
I realized that there are situations where on macos with apple silicon there may be versions of libraries compiled for different architectures.

>"Note that, just like ctypes.util.find_library, we currently don't take into account the architecture: the first library found with a suitable name is used, even if it doesn't match the architecture of the Python interpreter in use "

The main problem is that first all this is looked for in /usr/local and only then goes to /opt/homebrew/lib/

If in /usr/local there is a version of the library for an unnecessary architecture, then everything crashes.

Therefore, in this fix I made it so that a list of all libraries that are suitable is generated. As mentioned above, Python does not have the ability to determine the library architecture, so I just try to load them and if successful, everything works.

I don't really like exception and pass, but these are necessary measures for everything to work. In addition, it does not change the logic of code execution

### DIFF
--- a/usb/libloader.py
+++ b/usb/libloader.py
@@ -84,6 +84,8 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
                       Defaults to ctypes.util.find_library if not given or
                       None.
     """
+    liblist = []
+
     if find_library is None:
         find_library = ctypes.util.find_library
 
@@ -103,7 +105,7 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
 
         libname = find_library(candidate)
         if libname:
-            return libname
+            liblist.append(libname)
 
         # On Apple Silicon, also check in `/opt/homebrew/lib`: homebrew patches
         # the Python interpreters it distributes to check that directory, but
@@ -111,9 +113,9 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
         if use_apple_silicon_workaround:
             libname = "/opt/homebrew/lib/" + candidate + ".dylib"
             if os.path.isfile(libname):
-                return libname
-
-    return None
+                liblist.append(libname)
+        
+    return liblist
 
 def load_library(lib, name=None, lib_cls=None):
     """Loads a library. Catches and logs exceptions.
@@ -179,16 +181,20 @@ def load_locate_library(candidates, cygwin_lib, name,
         else:
             raise NoLibraryCandidatesException(name)
     elif candidates:
-        lib = locate_library(candidates, find_library)
-        if lib:
-            _LOGGER.debug("%r found as %s", (name or candidates), lib)
-            if sys.platform == 'win32':
-                loaded_lib = load_library(lib, name, win_cls)
-            else:
-                loaded_lib = load_library(lib, name, others_cls)
-        else:
-            _LOGGER.error('%r could not be found', (name or candidates))
-            raise LibraryNotFoundException(name)
+        liblist = locate_library(candidates, find_library)
+        for lib in liblist:
+            try:
+                if lib:
+                    _LOGGER.debug("%r found as %s", (name or candidates), lib)
+                    if sys.platform == 'win32':
+                        loaded_lib = load_library(lib, name, win_cls)
+                    else:
+                        loaded_lib = load_library(lib, name, others_cls)
+                else:
+                    _LOGGER.error('%r could not be found', (name or candidates))
+                    raise LibraryNotFoundException(name)
+            except OSError:
+                pass
     else:
         raise NoLibraryCandidatesException(name)
 


usb/libloader.py
```diff
@@ -103,17 +105,17 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
...
         # other/stock interpreters don't know about it
         if use_apple_silicon_workaround:
             libname = "/opt/homebrew/lib/" + candidate + ".dylib"
             if os.path.isfile(libname):
-                return libname
-
-    return None
+                liblist.append(libname)
+        
+    return liblist
```
jonasmalacofilho, line 118: We can't change the return type of the existing API, as it would be a breaking change.

For a new API, a generator that `yield`s candidates might be a cleaner way to do it. But because choices like these, in a library as mature as this one, I would also prefer that the new API be internal only in the beginning.

That said, it should be fairly simple to re-implement the old, public, API on top of the new, possibly generator-based, one (avoiding duplication).

usb/libloader.py
```diff
@@ -179,16 +181,20 @@ def load_locate_library(candidates, cygwin_lib, name,
...
+            try:
+                if lib:
+                    _LOGGER.debug("%r found as %s", (name or candidates), lib)
+                    if sys.platform == 'win32':
+                        loaded_lib = load_library(lib, name, win_cls)
+                    else:
+                        loaded_lib = load_library(lib, name, others_cls)
+                else:
+                    _LOGGER.error('%r could not be found', (name or candidates))
+                    raise LibraryNotFoundException(name)
```
jonasmalacofilho, line 195: It seems that the case where no library candidate is found isn't properly handled: in your current implementation, `locate_library` would return an empty list, and the whole loop body would never execute, making these lines unreachable.

jonasmalacofilho (commented):
I understand the motivation. In fact, homebrew's patch is slightly different than what I implemented in #511 in this very aspect: they somewhat prioritize `/opt/homebrew/lib`, while I considered it a last resort. In any case, both approaches will eventually fail in some scenarios.[^1]

Just so we're on the same page: were you already affected by this issue in the real world?

[^1]: I suspect mine fails in more of the common scenarios, but is also more flexible and respects what the user places in `~/lib`.

jonasmalacofilho:
> Here is the output of the console command as it currently looks on my system:

I'm more interested to know whether you originally discovered the issue because something actually failed to find the right library.

---

I can see how the correct library may be missed with your current setup, but it's curious that you have a x86-64 library in `/usr/local/lib`... how did it get there?

(Also note that only non-Homebrew Python interpreters should be affected).

mikhail-ekzi:
> I'm more interested to know whether you originally discovered the issue because something actually failed to find the right library.

The original x86_64 libraries were installed by a tool for programming STM32 microcontrollers. Later, I decided to use a some kind of a Python script that had only PyUSB as a dependency. That's when this bug appeared. Diagnostics with the PYUSB_DEBUG=debug variable revealed that the problem was with the library architecture.

I downloaded the original version via pip, and there were no commits for Apple Silicon, so I was very surprised to see them in the Git repository :)
I rolled out a new version from Git and realized that the problem persisted. This is a short summary of the issue.

I think that the problem is essentially that crooked Java installers crammed the library in without paying attention to the MAC architecture. Perhaps rosette is to blame, I don't know where it stores the libraries and whether it can use them transparently.

```
~ ❯ pkgutil --file-info /usr/local/lib/libusb-1.0.dylib                                                                                                                    21:27:43
volume: /
path: /usr/local/lib/libusb-1.0.dylib

pkgid: com.st.pkg.st-stlink-server
pkg-version: 2.1.1-2
install-time: 1718291890
uid: 0
gid: 0
mode: 100755
~ ❯ pkgutil --file-info /usr/local/lib//libusb-1.0.0.dylib                                                                                                                 21:27:54
volume: /
path: /usr/local/lib//libusb-1.0.0.dylib

pkgid: com.st.pkg.st-stlink-server
pkg-version: 2.1.1-2
install-time: 1718291890
uid: 0
gid: 0
mode: 100755
```

jonasmalacofilho:
> The original x86_64 libraries were installed by a tool for programming STM32 microcontrollers.

Ah, that explains it, thanks.

> Perhaps rosette is to blame, I don't know where it stores the libraries and whether it can use them transparently.

I don't know about Rosetta specifically, but when I looked into the more general situation of having both x86 and Arm libraries installed without Homebrew-style hacks, I couldn't find a proper way to do it. I think the idea is for everyone to use multiarch libraries, but that still doesn't work when old software doesn't know about them and just installs an old x86 one. Plus it seems that building multiarch libraries is annoying. 

---

Anyway, improving this is definitively welcome. Can you take a look at my comments and see if you can come up with a new (internal, for now) API?

PR #527, opened 2025-01-06
When data_or_wLength is an array, keep the same object

kauwua:
In `v1.2.1`, when data_or_wLength was an array, a `TypeError` exception was triggered in `create_buffer` which means `_interop.as_array` was called and the same array object was used (modified in-place as expected).

In  `v1.3.0`, commit 3ea79b01af08c1e8b2e0096a266c2cbd347738db introduced a change in `create_buffer` that does not trigger the `TypeError` exception and instead creates a new array from the one passed in `data_or_wLength`.

This means a new array is created and receives data for IN transfers, not the array passed as argument.

### DIFF
--- a/usb/core.py
+++ b/usb/core.py
@@ -1078,9 +1078,9 @@ class Device(_objfinalizer.AutoFinalizedObject):
         number of bytes read.
         """
         try:
-            buff = util.create_buffer(data_or_wLength)
-        except TypeError:
             buff = _interop.as_array(data_or_wLength)
+        except TypeError:
+            buff = util.create_buffer(data_or_wLength)
 
         self._ctx.managed_open()
 


usb/core.py
```diff
@@ -1078,9 +1078,9 @@ def ctrl_transfer(self, bmRequestType, bRequest, wValue=0, wIndex=0,
         number of bytes read.
         """
         try:
-            buff = util.create_buffer(data_or_wLength)
-        except TypeError:
             buff = _interop.as_array(data_or_wLength)
+        except TypeError:
+            buff = util.create_buffer(data_or_wLength)
```
jonasmalacofilho, line 1083: Hey, thanks for the report.

I think that depending on `TypeError` being raised is fragile, and continuing to do so would be a mistake.

How about:

```py
if instance(data_or_wLength, int):
    buff = util.create_buffer(data_or_wLength)
else:
    buff = _interop.as_array(data_or_wLength)
```

kauwua: I agree that waiting for a `TypeError` to happen is not clean, however it might be faster than checking the type of `data_or_wLength` at each transfer. A good thing with the fix is that it puts `as_array` first in the `try` block, which means if someone gives an existing array there is little overhead, while keeping an overhead for the case where an array is created each time anyways (so there's already an overhead).

In case you prefer using `isinstance`, checking `data_or_wLength` against the abtract `numbers.Integral` type might be better(https://peps.python.org/pep-3141/), I checked that `isinstance(x, int)` does not work with Numpy (`isinstance(numpy.uint64(3), int)` doesn't work for instance) but `isinstance(x,numbers.Integral)` works and is more generic:

```python
import numbers
if isinstance(data_or_wLength, numbers.Integral):
    buff = util.create_buffer(data_or_wLength)
else:
    buff = _interop.as_array(data_or_wLength)
```

kauwua: I put the alternative fix with `isinstance` here https://github.com/pyusb/pyusb/pull/529

jonasmalacofilho: Optimizing for the case where a byte `array` is passed in and not breaking the support for numpy integers are both valid points, thanks! But considering them, I think maybe #530 is a more comprehensive fix/improvement than either this or #529.

Can you take a look at #530 and see if I missed anything, or if you have any other suggestions/concerns?

kauwua: I tested https://github.com/pyusb/pyusb/pull/530 and it's good for me

PR #531, opened 2025-02-04
Add Additional Library Location for Snapcraft Support

russellproud:
I have a project that is based on python-ecspos which depends on Pyusb. This project is built as a Snap which is then deployed onto an Ubuntu Core based system to communicate with a USB based receipt printer.

Libusb access is granted to a snap by specifying it as a stage-package in the ```snapcraft.yaml``` file per below example.

```
    stage-packages:
      - libusb-1.0-0
```

Libusb is then accessible via the path of ```$snap/usr/lib/{architecture}/libusb-1.0-0.so.0```. ```find_library``` currently does not find the backend and fails to access any USB device.

This patch adds a check if the PyUSB  is running as a SNAP, and if so, attempts to load the library from the $SNAP location.

This patch only caters to ```x86_64-linux-gnu``` architecture, though additional architectures can be added to the ```snap_architectures``` list.

### DIFF
--- a/usb/libloader.py
+++ b/usb/libloader.py
@@ -96,6 +96,9 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
             and find_library is ctypes.util.find_library
     )
 
+    snap_env = os.environ.get("SNAP")
+    use_snap_workaround = bool(snap_env) and os.path.isdir(snap_env)
+
     for candidate in candidates:
         # Workaround for CPython 3.3 issue#16283 / pyusb #14
         if use_dll_workaround:
@@ -112,6 +115,22 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
             libname = "/opt/homebrew/lib/" + candidate + ".dylib"
             if os.path.isfile(libname):
                 return libname
+
+        # On Snap, also check in the snap's library directories
+        # This is required for Snaps when using stage-packages for libusb support
+        if use_snap_workaround:
+            snap_architectures = [
+                "x86_64-linux-gnu",
+            ]
+
+            snap_root = os.environ.get("SNAP")
+
+            for snap_architecture in snap_architectures:
+                snap_candidate = os.path.join(
+                    snap_root, "usr", "lib", snap_architecture, candidate + ".so.0"
+                )
+                if os.path.isfile(snap_candidate):
+                    return snap_candidate
 
     return None
 
--- a/usb/libloader.py
+++ b/usb/libloader.py
@@ -123,11 +123,9 @@ def locate_library(candidates, find_library=ctypes.util.find_library):
                 "x86_64-linux-gnu",
             ]
 
-            snap_root = os.environ.get("SNAP")
-
             for snap_architecture in snap_architectures:
                 snap_candidate = os.path.join(
-                    snap_root, "usr", "lib", snap_architecture, candidate + ".so.0"
+                    snap_env, "usr", "lib", snap_architecture, candidate + ".so.0"
                 )
                 if os.path.isfile(snap_candidate):
                     return snap_candidate


jonasmalacofilho:
Are you doing anything different than other projects using system libraries from Python? Doesn't Snap bother to ship a (patched) Python interpreter which can actually find and use stage-package libraries?

I mean, it's a possibility, but then I'm a bit surprised this hasn't been properly addressed by the Snapcraft developers before.

Also, is a stage-package library the only way to load LibUSB from a snap?

russellproud:
This has been built based on the guides provided by Snapcraft. In saying that, their docs and their examples are limited and/or missing at times. Eg. their Python example app repository has no ```snapcraft.yaml``` in it.

But based on my experience and understanding, I am doing it as intended. The snap is being built with the base Python plugin provided by Snapcraft. I haven't dug into the source of this package, but all my testing failed until I overrode ```usb.libloader.locate_library``` with my own function to handle the snap path. The moment that was done, all worked as intended.

The only way to expose libusb to the snap is via the stage-package method. If it is not exposed there, it is not accessible to the snap as far as I understand it.

I'm not here nor there if you accept this PR, my issue is resolved, only did it to assist if others encounter in the future. Obviously, your call.

jonasmalacofilho:
Thanks for the additional context!

Can you also help me a bit with how snaps are built? Specifically, do you know in which cases the patch

https://github.com/canonical/snapcraft/blob/7ccefeabb8bf2b01b9456f97a22493546bd7f3a3/snap/snapcraft.yaml#L145

would or wouldn't be applied to the Python interpreted used by a project built like a snap?

The issue you had is clearly very common. At the same time, the necessary patch has been written. But, from what I could gather, some ways of building a snap don't use the patched Python interpreter.

russellproud:
Hey @jonasmalacofilho 

The commit here https://github.com/canonical/snapcraft/blob/7ccefeab/snap/snapcraft.yaml#L145 is a snap itself.

That yaml file is them building the Snapcraft Snap, which is published here https://snapcraft.io/snapcraft

So my reading of that is, they're patching themselves in their own snap build, which means the underlying python layer isn't patched.

jonasmalacofilho:
Note to self: understand why some projects using `staged-packages: python3-usb` supposedly just work.

In particular, no issue has been reported with the liquidctl snap, which depends on finding LibUSB for all actions, and just does that. Maybe `confinement` or the `python` plugin are relevant to the issue?

(For others reading this: I'm also the maintainer of the liquidctl upstream project. So I've seen reports that the liquidctl snap has some issues, but none of them are related to [not] finding LibUSB).

russellproud:
I think I misread your previous message. I thought you were pointing to the snapcraft commit thinking it was the python library. 

If I get some time today or tomorrow, I'll test a few alternative approaches to defining the snap and see if any result in successful access to libusb without modifications.

jonasmalacofilho:
> Also, quick q, do you have the snapcraft.yaml for liquidctl? I can't see it on the repo.

It was created by someone else: https://github.com/degville/liquidctl-snap/blob/main/snap/snapcraft.yaml

russellproud:
Ok, I think I see why. I'll confirm in a few hrs.

I suspect the python3-usb stage package is the reason.

My approach has been to install pyusb via pip (done by snapcraft at build time) and expose libusb as a stage package.

Python3-usb is an Canonical packaged version of pyusb. It exposes libusb I believe as part of it. Something may be different in the package or the approach vs installing 

Will test and confirm.

russellproud:
OK, I adjusted my snap to use python3-usb as the stage package vs libusb-1.0-0.

Rebuilt and redeployed onto an Ubuntu Core device. It fails to access the USB device. 

So in my instance, it's still failing when using python3-usb and libusb as stage packages.

If there's anything else specific, I can add, let me know.

Appreciate the convo and the package!

jonasmalacofilho:
@russellproud, thanks for the additional test.

Your pull request looks good as is. Obviously, having too many such workarounds here is not ideal, so I'll give myself a few more days before merging it... maybe I'll discover _something_ in the meantime.

Either way, I don't think there's anything else you need to do here. Really, thanks for the patch, as well (again) for the discussion and tests.

PR #534, opened 2025-08-04
Do not require wheel for building

hroncok:
- current version of setuptools (70.1+) does not need wheel at all
- older versions of setuptools would fetch wheel when building wheels (but not sdists)


jonasmalacofilho:
Looks reasonable. But if older versions of setuptools would fetch it anyway, why was it recommended to explicitly require `wheel` in the docs (at the time) and PEP 518?

jonasmalacofilho:
All right then. Thanks for the PR.

